PR #2, opened 2020-08-05
Adds basic crate structure

tarikeshaq:
Adds basic project structure

### DIFF
--- a/experiments/Cargo.toml
+++ b/experiments/Cargo.toml
@@ -15,3 +15,10 @@ crate-type = ["lib"]
 
 [dependencies]
 anyhow = "1"
+serde = "1"
+serde_derive = "1"
+log = "0.4"
+# Viaduct locked to version 61.0.10 of Application Services
+viaduct = { git = "https://github.com/mozilla/application-services",  rev = "2478bcf2b48d1867b01e8b7df4f86a69d564d49a"}
+thiserror = "1"
+
--- a/experiments/src/lib.rs
+++ b/experiments/src/lib.rs
@@ -2,9 +2,17 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use anyhow::{anyhow, Result};
+pub mod error;
+mod evaluator;
+pub use error::*;
+mod http_client;
+mod matcher;
+mod persistence;
 
-#[derive(Debug, Clone)]
+use anyhow::{anyhow, Result};
+use serde_derive::*;
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Experiment {
     slug: String,
 }
--- a/experiments/src/matcher.rs
+++ b/experiments/src/matcher.rs
@@ -1,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! This module defines all the information needed to match a user with an experiment
+//! this module should also include a `match` function of some sort that does the matching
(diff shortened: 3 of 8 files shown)


experiments/Cargo.toml
```diff
@@ -15,3 +15,9 @@ crate-type = ["lib"]
 
 [dependencies]
 anyhow = "1"
+serde = "1"
+serde_derive = "1"
+log = "0.4"
+viaduct = { git = "https://github.com/mozilla/application-services" }
```
vladikoff, line 21: can we add a GIT SHA here? Lock to a specific release maybe?

experiments/src/buckets.rs
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! This might be where the bucketing logic can go
```
vladikoff, line 5: @tarikeshaq while porting this to Rust, does it make sense to call this file `evaluator`, to mimic what https://github.com/mozilla-mobile/android-components/blob/master/components/service/experiments/src/main/java/mozilla/components/service/experiments/ExperimentEvaluator.kt does (bucketing and more)?

experiments/src/matcher.rs
```diff
@@ -0,0 +1,36 @@
...
+//! This module defines all the information needed to match a user with an experiment
+//! this module should also include a `match` function of some sort that does the matching
+//! it has two main types, the `matcher` retrieved from the server, and the `AppContext`
+//! from the client
+//! Note: This could be where the logic to evaluate the filter_expressions lies
+use serde_derive::*;
+
+#[derive(Deserialize, Serialize, Debug, Clone, Default)]
+pub struct Matcher {
+    pub app_id: Option<String>,
```
vladikoff, line 14: @pdehaan do you have a list of fields that Firefox Desktop implementation provides for Experiment context? Are we missing anything?

experiments/src/metrics_manager.rs
```diff
@@ -0,0 +1,6 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
vladikoff, line 1: We can remove this file for now I think

experiments/src/buckets.rs
travis79: I second this, if the plan is to have the bucketing and matching evaluation in this module then 'evaluator' makes more sense to me.

experiments/src/lib.rs
```diff
@@ -2,9 +2,17 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+mod buckets;
```
travis79, line 5: This should be `evaluator` now, right?

experiments/src/matcher.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub app_display_version: Option<String>,
+    pub app_min_version: Option<String>,
+    pub app_max_version: Option<String>,
+    pub locale_language: Option<String>,
+    pub locale_country: Option<String>,
```
travis79, line 19: service/experiments splits locale into country and language but Glean records this as the combined format (en-US, en-GB, sp-MX, de-DE, etc..)  I propose combining these into `locale` and going with the same format we use in telemetry to make analysis easier for the data science folks.

experiments/src/lib.rs
tarikeshaq: Yup! Good catch 🤦

experiments/src/matcher.rs
travis79: I believe Desktop is using JEXL for their targeting/matching, and the only docs I found on that [are here](https://firefox-source-docs.mozilla.org/browser/components/newtab/content-src/asrouter/docs/targeting-guide.html?highlight=jexl%20targeting)

PR #3, opened 2020-08-05
Adds nimbus schema and client for experiements

tarikeshaq:
The client probably will change to test with a real server, but added enough to test retrieving a valid nimbus experiment (and converting it into a Rust struct)

Also, note that the majority of the `lines` in this PR are the Cargo.lock lines


experiments/src/lib.rs
```diff
@@ -44,3 +40,67 @@ impl Experiments {
...
+pub struct BucketConfig {
+    pub randomization_unit: RandomizationUnit,
+    pub namespace: String,
+    pub start: u32,
+    pub count: u32,
+    pub total: u32,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
```
travis79, line 101: According to [the experiment schema](https://github.com/mozilla/nimbus-shared/blob/8bbb142b6cfdebc285b29d24e37fadc8bef0ee4c/types/experiments.ts#L60), it looks like `userId` isn't snake case.  I'm not sure if this is a bug in the schema or by design, but if we snake case here it probably wouldn't match the schema as it is defined right now.

tarikeshaq: Wow that's a good catch! Thanks @travis79!!

travis79: FYI: I asked in #nimbus-project on Slack to get more info on this

PR #4, opened 2020-08-05
Adds basic persistence.

tarikeshaq:
Uses `rkv` for persistence. 

This uses `rkv`'s `lmdb` backend, which is [unstable](https://github.com/mozilla/rkv), I'll change it to `SafeMode` on a follow-up based on the feedback I get here.

### DIFF
--- a/experiments/Cargo.lock
+++ b/experiments/Cargo.lock
@@ -16,6 +16,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6b602bfe940d21c130f3895acd65221e8a61270debe89d628b9cb4e3ccb8569b"
 
 [[package]]
+name = "arrayref"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4c527152e37cf757a3f78aae5a06fbeefdb07ccc535c980a3208ee3060dd544"
+
+[[package]]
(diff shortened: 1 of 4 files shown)


mdboom:
> Uses `rkv` for persistence.
> 
> This uses `rkv`'s `lmdb` backend, which is [unstable](https://github.com/mozilla/rkv), I'll change it to `SafeMode` on a follow-up based on the feedback I get here.

FWIW, Glean also uses the lmdb backend.  We've run into a few issues, but they were mainly around the lack of thread safety in some of the setup/teardown parts.  In practice, that hasn't been an issue because these are basically statics during the life of the application -- we only hit it in testing where we create/destroy a lot more of these things.

@badboy: I think you had some other thoughts that were keeping Glean away from safe mode...

badboy:
I talked with Victor a while ago about that.
Safe mode is essentially an in-memory BTreeMap, synced to disk once in a while.
As we haven't experienced the issues with Rkv/LMDB in Glean that popped up on Desktop we are staying on the LMDB mode for now. This will probably reevaluated once we land some proper code in mozilla-central (and that requires additional work if we actually need to migrate databases).

Once we bundle nimbus & Glean we _probably_ need to upgrade the rkv crate inside Glean to avoid any duplication or linker issues. Luckily newer versions offer both modes, so nimbus can use the one and we the other, given that those are separate database anyway.

PR #5, opened 2020-08-06
Adds custom config for initializing experiments

tarikeshaq:
Adds some config parameters that can be passed in for debugging

### DIFF
--- a/experiments/src/lib.rs
+++ b/experiments/src/lib.rs
@@ -5,34 +5,45 @@
 pub mod error;
 mod evaluator;
 pub use error::*;
+mod config;
 mod http_client;
 mod matcher;
 mod persistence;
+mod uuid;
 
+use ::uuid::Uuid;
 use anyhow::{anyhow, Result};
 use chrono::{DateTime, Utc};
+pub use config::Config;
+pub use matcher::AppContext;
 use serde_derive::*;
+use std::path::Path;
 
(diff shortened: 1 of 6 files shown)


experiments/src/lib.rs
```diff
@@ -10,30 +10,52 @@ mod matcher;
...
 impl Experiments {
-    pub fn new() -> Self {
+    pub fn new<P: AsRef<Path>>(
+        app_context: AppContext,
+        _db_path: P,
+        config: Option<Config>,
+    ) -> Self {
         let resp = vec![];
-        Self { experiments: resp }
+        let uuid = match config {
```
vladikoff, line 50: @tarikeshaq is it possible to move out UUID generation in its own file?

vladikoff: Just like we have for AppContext.

experiments/src/lib.rs
```diff
@@ -10,30 +10,52 @@ mod matcher;
...
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Experiment {
     slug: String,
 }
+
+/// Optional custom configuration
+/// Could be used for debugging, or to change the server the SDK will communicate with
+/// Also includes an optional `uuid` for debugging/testing purposes
+/// The `uuid` would typically be retrieved from persisted storage
+pub struct Config {
```
vladikoff, line 26: Can we move this in its own file as well? (just in case config grows and we can document more of its properties)

PR #7, opened 2020-08-06
Fix up CI

vladikoff:


pdehaan:
Do we want to add Rust linting w/ **clippy** like in https://github.com/mozilla/uniffi-rs/blob/main/.circleci/config.yml?

vladikoff:
> Do we want to add Rust linting w/ clippy 

Yeah later on 👍 , I'll file a bug  https://jira.mozilla.com/browse/SYNC-1625

PR #8, opened 2020-08-10
[Don't Merge] Runs example against stage

tarikeshaq:
Runs the example against the stage environment. 

`Don't merge` because some experiments on the stage environment don't have some required fields, and to make the example run I changed up our schema to allow for that


vladikoff:
The work in https://github.com/mozilla/nimbus-sdk/pull/16 already does this 👍

PR #9, opened 2020-08-10
First pass at adding jexl-rs

tarikeshaq:
First pass at adding jexl evaluation.

`jexl-rs` still needs a bit more work before it's ready though.

Marking as a draft until we put `jexl-rs` into crates.io or (decide to just point it to GitHub)

### DIFF
--- a/experiments/src/evaluator.rs
+++ b/experiments/src/evaluator.rs
@@ -9,8 +9,10 @@
 
 //! TODO: Implement the bucketing logic from the nimbus project
 
+use crate::error::{Error, Result};
+use crate::matcher::AppContext;
+use jexl_eval::eval_in_context;
 use serde_derive::*;
-
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
 pub struct Bucket {}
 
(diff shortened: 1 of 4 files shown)


experiments/src/matcher.rs
```diff
@@ -25,7 +25,7 @@ pub struct Matcher {
 #[derive(Deserialize, Serialize, Debug, Clone, Default)]
 pub struct AppContext {
     pub app_id: Option<String>,
-    pub app_version: Option<String>,
+    pub app_version: Option<u32>,
```
vladikoff, line 28: I am not sure if this will work, the app versions we have can be something like 79.4. I am also worried that they might also just be strings everywhere in the experimenter platform

tarikeshaq: Ya that's a good point. I'll change it back for now! Transforms will make the comparison here possible with string versions, but we can worry about that later.

experiments/src/evaluator.rs
```diff
@@ -21,4 +23,49 @@ impl Bucket {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_targeting() {
+        // Here's our valid jexl statement
+        let expression_statement =
+            "app_id == '1010' && ( app_version == '4.4' || locale == \"En-US\")";
```
vladikoff, line 56:
```suggestion
            "app_id == '1010' && ( app_version == '4.4' || locale == \"en-US\")";
```

experiments/src/evaluator.rs
```diff
@@ -21,4 +23,49 @@ impl Bucket {
...
+        // A valid context
+        let ctx = AppContext {
+            app_id: Some("1010".to_string()),
+            app_version: Some("4.4".to_string()),
+            locale: Some("En-US".to_string()),
```
vladikoff, line 61:
```suggestion
            locale: Some("en-US".to_string()),
```

PR #11, opened 2020-08-18
[WIP] First pass at adding bucketing logic

tarikeshaq:
#### What is this?
This is a first pass at adding the required bucketing logic. It implements two high-level functions
- `bucket_sample`: that decides if a user should enroll given an input (usually the id, along with the namespace) (ref in desktop here)
- `choose_branch`: Randomly decides on a branch for a user given ratios of the branches.

It also adds a `sampling` module that deals with all the hashing/sampling.

#### What else do we need to add?
- [ ] The actual logic to enroll in experiments based on this logic and the `bucket_config`
- [ ] Tests for both functions
- [ ] An integration test that shows full experiment enrollment flow
- [ ] Tests for the sampling functions
- [ ] Investigating anything else that needs to be added as a first pass + cleaning this up


experiments/src/evaluator.rs
```diff
@@ -23,13 +25,75 @@ impl Bucket {
...
+///
+/// - `input` What will be hashed and matched against the range of the buckets
+/// - `start` the index of the bucket to start checking
+/// - `count` then number of buckets to check
+/// - `total` The total number of buckets to group inputs into
+///
+/// # Returns:
+///
+/// Returns true if the hash generated from the input belongs within the range
+// otherwise false
```
vladikoff, line 45: format nit: `///` ?

experiments/src/evaluator.rs
```diff
@@ -23,13 +25,75 @@ impl Bucket {
     }
 }
 
+/// Sample by splitting the input space into a series of buckets, checking
+/// if the given input is in a range of buckets
+///
+/// The range to check is defined by a start point and length, and can wrap around
+/// the input space. For example, if there are 100 buckets, and we ask to check 50 buckets
+/// starting from bucket 70, then buckets 70-99 and 0-19 will be checked
```
vladikoff, line 33: "then buckets 70-99 and 0-19 will be checked" TIL

experiments/src/evaluator.rs
```diff
@@ -23,13 +25,75 @@ impl Bucket {
...
+}
+
 /// Checks if the client is targeted by an experiment
 /// This api evaluates the JEXL statement retrieved from the server
 /// against the application context provided by the client
 ///
 /// # Arguments
 /// - `expression_statement`: The JEXL statement provided by the server
-/// - `ctx`: The application context provided by the client
+/// - `ctx`: The application context provided by the client∏
```
vladikoff, line 96: ∏_∏

experiments/src/sampling.rs
```diff
@@ -0,0 +1,130 @@
...
+    let padding = vec!['0'; HASH_LENGTH as usize];
+    let res = padding
+        .into_iter()
+        .chain(multiplied.chars())
+        .collect::<String>();
+    Ok(res)
+}
+
+#[cfg(test)]
+mod tests {
```
vladikoff, line 92: thank you for adding these

vladikoff:
@tarikeshaq does it make sense to land some of this in a smaller PR and follow up with items that are on your list?

PR #12, opened 2020-08-20
Updates application context and targeting attributes

travis79:
This refactors the list of application context attributes that are used for matching purposes to be more in line with the `client_info` section collected by Glean telemetry in order to ease the burden on analysis.

Also expands the evaluator tests a bit.


experiments/src/matcher.rs
```diff
@@ -15,20 +15,30 @@ pub struct Matcher {
...
+    pub app_max_build: Option<String>,
+    pub architecture: Option<String>,
     pub device_manufacturer: Option<String>,
     pub device_model: Option<String>,
-    pub regions: Vec<String>,
+    pub locale: Option<String>,
+    pub os: Option<String>,
+    pub os_version: Option<String>,
+    pub android_sdk_version: Option<String>,
     pub debug_tags: Vec<String>,
```
vladikoff, line 28: Thank you for updating this list!

PR #15, opened 2020-08-21
Adds basic uniffi integration

tarikeshaq:


experiments/Cargo.toml
```diff
@@ -29,7 +29,12 @@ jexl-eval = "0.1.6"
 uuid = { version = "0.8", features = ["serde", "v4"]}
 sha2 = "0.9"
 hex = "0.4"
+# We use a fixed version of ffi-support to resolve the conflict between viaduct and uniffi in the versions of ffi_support they use
+ffi-support = "0.4.2"
+uniffi = { git = "https://github.com/mozilla/uniffi-rs", branch = "main" }
```
vladikoff, line 34: a bit dangerous to lock to a `main` branch, I am not sure if we should lock it to a GIT SHA at this point or not.

tarikeshaq: opened https://github.com/mozilla/uniffi-rs/issues/253 to publish the uniffi crates to crates.io, for the time being, I can lock it to the latest commit

travis79:
One question, when we do generate the other language bindings, will those get directories and project files generated also?

experiments/Cargo.toml
```diff
@@ -29,7 +29,12 @@ jexl-eval = "0.1.6"
 uuid = { version = "0.8", features = ["serde", "v4"]}
 sha2 = "0.9"
 hex = "0.4"
+# We use a fixed version of ffi-support to resolve the conflict between viaduct and uniffi in the versions of ffi_support they use
```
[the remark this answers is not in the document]
tarikeshaq: Hmmm looks like this issue is gone.... Not 100% sure why it's gone, but my guess is the Cargo.lock is helping us out here 😄

I don't think we should need to match the versions and it's weird that the issue showed up in the first place, since viaduct uses version "0.4" and uniffi uses "~0.4.2". I would naively expect cargo to know to pick "0.4.2"... Anywho, not an issue anymore

tarikeshaq:
@travis79 Right now `uniffi` will only generate the files (i.e if we are generating the kotlin bindings, it will generate 1 `.kt` file)

If you take a look at https://github.com/mozilla/application-services/pull/3491/files#diff-abbb69499a6d2ef1a59ffc69b6ef367eR129-R139 @eoger used gradle tasks to generate the binding into an android project

We could do something similar here, or even generate the bindings and check them in.

But either way, I think we would need to create the android structure ourselves. (I hope I didn't misunderstand your question 😅)

travis79:
> @travis79 Right now `uniffi` will only generate the files (i.e if we are generating the kotlin bindings, it will generate 1 `.kt` file)
> 
> If you take a look at [mozilla/application-services#3491 (files)](https://github.com/mozilla/application-services/pull/3491/files#diff-abbb69499a6d2ef1a59ffc69b6ef367eR129-R139) @eoger used gradle tasks to generate the binding into an android project
> 
> We could do something similar here, or even generate the bindings and check them in.
> 
> But either way, I think we would need to create the android structure ourselves. (I hope I didn't misunderstand your question 😅)

Nope, that's exactly what I wanted to know!

experiments/Cargo.toml
```diff
@@ -29,7 +29,10 @@ jexl-eval = "0.1.6"
 uuid = { version = "0.8", features = ["serde", "v4"]}
 sha2 = "0.9"
 hex = "0.4"
+uniffi = { git = "https://github.com/mozilla/uniffi-rs", rev = "686a8ec672cb6b86a6de920c43ff2d5678afd84a" }
 
+[build-dependencies]
+uniffi_build = { git = "https://github.com/mozilla/uniffi-rs", rev = "686a8ec672cb6b86a6de920c43ff2d5678afd84a", features=["builtin-bindgen"] }
```
rfk, line 35: The `builtin-bindgen` here will cause cargo to build a local copy of the `uniffi-bindgen` tool and use that for generating the rust scaffolding.

An alternative is to have developers install `uniffi-bindgen` on their system, remove the `builtin-bindgen` feature, and let this use the system-provided tool. The potential advantage of this is that it's easier to ensure the kotlin bindings are generated using the same version of the tool; the downside is a less convenient developer experience.

rfk (approved):
Looks good to me!

I agree with what (I think) is said in the comments about, about also trying to incorporate the generation of the Kotlin bindings sooner rather than later. It's important that the Kotlin bindings are generated with the same version of uniffi as the rust scaffolding, and it's not obvious from this PR how you'll enforce that.

experiments/Cargo.toml
vladikoff: " the downside is a less convenient developer experience." is this just because they have to build it again? it didn't seem to take that long when I tried.

tarikeshaq: The inconvenience is probably that developers have to ensure that they have `uniffi-bindgen` installed and in their `PATH` environment variable (a `cargo install` suffices here).
>  It's important that the Kotlin bindings are generated with the same version of uniffi as the rust scaffolding, and it's not obvious from this PR how you'll enforce that.

Ya I agree it's super important. When I add the kotlin bits I'll make sure we have a nice way to enforce that (Removing the `builtin-bindgen` seems like the most straight forward way)

rfk: Right, yeah, it's an inconvenience in that it adds another thing to the list of prerequisites you have to install on your machine in order to build the project.

PR #16, opened 2020-08-24
Skips invalid experiments recieved from the server

tarikeshaq:
This PR allows us to query experiments from [stage](https://settings.stage.mozaws.net/v1/buckets/main/collections/messaging-experiments/records) that abide by the nimbus schema, and omitting experiments that do not abide by the schema (with a minor detail, the `proposedDuration` is missing from ***all*** experiments in stage, so I made it optional (although the schema said it's required) and added a TODO there.

### DIFF
--- a/experiments/examples/experiment.rs
+++ b/experiments/examples/experiment.rs
@@ -5,7 +5,24 @@
 use anyhow::Result;
 use experiments::{AppContext, Experiments};
 fn main() -> Result<()> {
-    let exp = Experiments::new(AppContext::default(), "../target/mydb", None).unwrap();
-    exp.get_active_experiments();
+    use env_logger::Env;
+    // We set the logging level to be `warn` here, meaning that only
+    // logs of `warn` or higher will be actually be shown, any other
+    // error will be omitted
+    // To manually set the log level, you can set the `RUST_LOG` environment variable
+    // Possible values are "info", "debug", "warn" and "error"
+    // Check [`env_logger`](https://docs.rs/env_logger/) for more details
+    env_logger::from_env(Env::default().default_filter_or("warn")).init();
+    viaduct_reqwest::use_reqwest_backend();
+    let exp = Experiments::new(
+        "messaging-experiments".to_string(),
+        AppContext::default(),
+        "../target/mydb",
+        None,
+    )
+    .unwrap();
+    exp.get_active_experiments()
+        .iter()
+        .for_each(|e| println!("Experiment: {}", e.slug));
     Ok(())
 }
--- a/experiments/src/http_client.rs
+++ b/experiments/src/http_client.rs
@@ -7,9 +7,13 @@
 //! Working on the real Nimbus schema.
 
 use super::Experiment;
-use anyhow::Result;
+use crate::config::Config;
+use crate::error::{Error, Result};
 use url::Url;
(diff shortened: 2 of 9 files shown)


experiments/src/http_client.rs
```diff
@@ -7,16 +7,76 @@
...
+///
+/// This is needed to filter out experiments that do not abide by the
+/// schema. For production, we would probably need to remove this
+/// and ensure that the server contains *only* valid experiments
+fn skip_invalid<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+    T: serde::Deserialize<'de>,
+{
+    struct SkipInvalid<T>(PhantomData<T>);
```
vladikoff, line 43: let's log when we find a badly formatted experiment

experiments/src/http_client.rs
```diff
@@ -7,16 +7,79 @@
...
+            loop {
+                match seq.next_element::<T>() {
+                    Ok(inner) => {
+                        if let Some(inner) = inner {
+                            vec.push(inner);
+                        } else {
+                            break;
+                        }
+                    }
+                    Err(e) => log::warn!("Malformed experiment found! Error: {}", e),
```
vladikoff, line 74: could we please include the name of the experiment here as well?

experiments/src/http_client.rs
```diff
@@ -7,16 +7,79 @@
...
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: SeqAccess<'de>,
+        {
+            let mut vec = Vec::new();
+            // NOTE: In here we ignore any errors that appear in
+            // the deserialization process and instead
+            // skip any invalid elements
+            // In the future (when we have a source of *only* valid experiments)
```
vladikoff, line 63: I think in the future there will probably be malformed experiments. should we tweak the comment?

experiments/examples/experiment.rs
```diff
@@ -5,7 +5,13 @@
 use anyhow::Result;
 use experiments::{AppContext, Experiments};
 fn main() -> Result<()> {
-    let exp = Experiments::new(AppContext::default(), "../target/mydb", None);
-    exp.get_experiments();
+    use env_logger::Env;
+
+    env_logger::from_env(Env::default().default_filter_or("warn")).init();
```
vladikoff, line 10: could you please add a comment here about the usage for the ENV_LOGGER, the docs are in https://docs.rs/env_logger/0.7.1/env_logger/ but it would be useful to have a quick note about how to tweak the log level right in here. I think it would be useful once we have more trace logging in the SDK

experiments/src/http_client.rs
tarikeshaq: Hmmm I'll take a look, may not be easy since all we receive back from `next_element` is the error. Should be possible with more playing around thou

vladikoff: yea, feel free to time-box this as well. It's not critical, but makes it harder to find the broken experiment. We can also log it in a different statement afterwards, not sure of that helps.

tarikeshaq: Figured out a way around this 😄

vladikoff:
@tarikeshaq Awesome!! Thanks for the updates!

PR #18, opened 2020-08-26
Adds kotlin bindings generation

tarikeshaq:
Initial work to generate the kotlin bindings

The most meaningful work here is in the `build.gradle`, which itself was mostly copied from our work at Application Services. Almost everything else was generated by Andriod Studio


android/src/test/java/org/mozilla/nimbus/ExampleUnitTest.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test
+
+import org.junit.Assert.*
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+class ExampleUnitTest {
```
vladikoff, line 12: @tarikeshaq i think it's also worth adding a README for the `android` directory. Mostly a HOWTO build locally I think

travis79: Yes, any environment requirements or setup instructions would be useful.

android/build.gradle
```diff
@@ -0,0 +1,224 @@
...
+            }
+            maven {
+                name = "projectBuildDir"
+                url "file://${project.buildDir}/maven"
+            }
+        }
+    }
+}
+
+ext.configurePublish = { jnaForTestConfiguration = null ->
```
tarikeshaq, line 171: This should probably go into a `publish.gradle` like we do in other projects

tarikeshaq: I'm not too familiar with the details of the publishing,

But we'd need to add a .buildconfig to retrieve the values that I have hard coded here

(i.e the name of the package, the version etc)

android/build.gradle
```diff
@@ -0,0 +1,224 @@
...
+    def theGroupId = "uniffi"
+    def theArtifactId = "nimbus"
+    def theDescription = "Nimbus SDK"
+
+    task sourcesJar(type: Jar) {
+        from android.sourceSets.main.java.srcDirs
+        classifier = 'sources'
+    }
+
+    publishing {
```
tarikeshaq, line 181: This was copied from the Application services `publish.gradle` I'm not too familiar with the details

android/build.gradle
```diff
@@ -0,0 +1,224 @@
...
+                    }
+
+                }
+            }
+        }
+    }
+
+    task checkMavenArtifacts
+
+    publishing.publications.withType(MavenPublication).each { publication ->
```
tarikeshaq, line 215: This should be removed

android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="uniffi.nimbus">
```
tarikeshaq, line 2: This should be `org.mozilla.nimbus` but keeping it like this because the package name of the generated bindings is `uniffi.nimbus` by default

.circleci/config.yml
```diff
@@ -9,6 +9,9 @@ jobs:
       - run:
           name: Version information
           command: rustc --version; cargo --version; rustup --version
+      - run:
+          name: Install uniffi
+          command: git clone https://github.com/mozilla/uniffi-rs.git; cargo install --path uniffi-rs/uniffi_bindgen
```
tarikeshaq, line 14: A placeholder till https://github.com/mozilla/uniffi-rs/issues/253 is resolved.

It can then be replaced with 

```
cargo install uniffi_bindgen
```

tarikeshaq:
Gonna drop a comment here before I forget, work here should be OK for publishing to maven local and working with that. There are some details here to be figured out about fully publishing this though

PR #23, opened 2020-09-16
Install uniffi from crates.io in CircleCI

rfk:
It's important that the version installed here matches the version used in the `Cargo.toml` of the experiments crate.

I think this should fix the issue from https://github.com/mozilla/nimbus-sdk/pull/22, closing https://github.com/mozilla/application-services/issues/3583.


vladikoff (approved):
Thank you!!

rfk:
I filed https://github.com/mozilla/uniffi-rs/issues/284 to follow up on the developer experience here.

PR #24, opened 2020-09-22
Replace RandomizationUnit with a plain Uuid.

mhammond:
I think there was some confusion around the discussion about how to choose
the Uuid used as the RandomizationUnit. We might need an enum later if we
do need to choose between different Uuids based on the experiment, but we
can add that once the requirements are clearer.

### DIFF
--- a/experiments/src/evaluator.rs
+++ b/experiments/src/evaluator.rs
@@ -128,7 +128,7 @@ pub(crate) fn targeting(expression_statement: &str, ctx: AppContext) -> Result<b
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{BucketConfig, ExperimentArguments, FeatureConfig, Group, RandomizationUnit};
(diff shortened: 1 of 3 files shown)


rfk (commented):
This doesn't feel right to me, but I also realized I was using some imprecise terminology in our slack convo earlier which may have contributed to confusion.

IIUC the `BucketConfig` is data that we fetch from the server, describing how to allocate users to the different variants in the experiment. This config will by necessity be the same for all users. It says *what property we want to use to bucket users* but can't tell us the actual value of that property for a specific user.

I would expect the specific id value to either be provided by the application as part of [`AppContext`](https://github.com/mozilla/nimbus-sdk/blob/main/experiments/src/matcher.rs#L33), or created internally by the crate and persisted in local storage.

It may be that we only support a single type of id at this stage, and so we don't need to be told a `randomization_unit` value at all by the server. But we still need to accept the value from somewhere.

mhammond:
Right, thanks - the server does send it, and we don't use it as input to bucketing function, so yeah... :)

PR #25, opened 2020-09-22
[SYNC-1786] Rename "experiments" to "nimbus"

travis79:
This renames "experiments" to "nimbus" to be less confusing with the existing Android Component.


nimbus/Cargo.toml
```diff
@@ -1,13 +1,13 @@
 [package]
-name = "experiments"
+name = "nimbus-sdk"
```
travis79, line 2: I chose this, just in case we wanted to put a crate up on crates.io as this name wasn't taken (nimbus was).

mhammond:
The "experiments" name certainly confused me, so I think this is a better name. Does anyone know why we have this directory layout at all? ie, why not just have `Cargo.toml` and `src` in the root of the tree? Then the name makes perfect sense and any other name would be self-evidently crazy :)

rfk:
> why not just have Cargo.toml and src in the root of the tree?

I don't know, but I think I'd personally prefer this layout as well, since it's similar to how the appservices rust components are structured.

nimbus/src/nimbus.idl
```diff
@@ -34,7 +34,7 @@ enum Error {
     "RequestError", "ResponseError", "InvalidExperimentResponse"
 };
 
-interface Experiments {
+interface Nimbus {
```
rfk, line 37: I don't want to bikeshed this too much, but in general I'm not a fan of cute project codenames leaking into APIs, because they don't help you understand the point of the code unless you're already familiar with it. I think the ship has already sailed on this one though :-)

Would something like `NimbusClient` or similar be more helpful here with building the mental model; knowing that "this is a client for "Nimbus", whatever that is" seems marginally more useful than knowing that "this is a Nimbus, whatever that is".

I won't object either way though.

rfk (approved):
This is a marked improvement, thank you! I think we can safely land this and tweak the file layout further over time if we want to.

nimbus/src/nimbus.idl
travis79: I really have no objections to naming this NimbusClient (I was also thinking something like NimbusSDK). I'm all for descriptive naming and that would save having to define alias in the A-C component if we also wanted to call it by its "cute" name ;).  I'll go ahead and change this to NimbusClient in this PR since you had a similar issue with this.

PR #28, opened 2020-09-25
WIP - init viadict

mhammond:
This "works" insofar as we make a request to the stage server.


rfk:
I had a brief poke at this, but it seems to fail for me in the same way it fails for you (which is good I guess!). FWIW I wouldn't be opposed to dragging along the protobuf plugin if it helps us get to something that works short-term.

travis79:
I'll check this out and give it a go today 🤞

travis79:
Here's my go at getting this working for you, hope it helps: https://github.com/mhammond/nimbus-sdk/pull/1

mhammond:
Abandoning this for now in favor of the nimbus-megazord branch.

PR #30, opened 2020-09-29
Update to track latest nimbus schema

rfk:
The experiment schema defined over in [nimbus-shared](https://github.com/k88hudson/nimbus-shared/blob/fb9bacac84fedd6c268621add00966775b6c9dd2/types/experiments.ts#L77) has had a few modifications recently, which IMHO make it substantially better for our purposes. I was going through the details and figured I might as well update the rust code to track it while I was in there.

This is WIP only because I don't know whether we need to wait for these schema changes to get deployed to a server somewhere, ref https://jira.mozilla.com/browse/SYNC-1794

### DIFF
--- a/nimbus/src/evaluator.rs
+++ b/nimbus/src/evaluator.rs
@@ -48,7 +48,7 @@ impl Bucket {
 pub fn filter_enrolled(id: &Uuid, experiments: &[Experiment]) -> Result<Vec<EnrolledExperiment>> {
     let mut res = Vec::with_capacity(experiments.len());
     for exp in experiments {
-        let bucket_config = exp.arguments.bucket_config.clone();
+        let bucket_config = exp.bucket_config.clone();
         if sampling::bucket_sample(
             vec![id.to_string(), bucket_config.namespace],
             bucket_config.start,
@@ -56,12 +56,10 @@ pub fn filter_enrolled(id: &Uuid, experiments: &[Experiment]) -> Result<Vec<Enro
             bucket_config.total,
         )? {
             res.push(EnrolledExperiment {
-                slug: exp.arguments.slug.clone(),
-                user_facing_name: exp.arguments.user_facing_name.clone(),
-                user_facing_description: exp.arguments.user_facing_description.clone(),
-                branch_slug: choose_branch(&exp.arguments.slug, &exp.arguments.branches, id)?
-                    .clone()
-                    .slug,
+                slug: exp.slug.clone(),
+                user_facing_name: exp.user_facing_name.clone(),
+                user_facing_description: exp.user_facing_description.clone(),
+                branch_slug: choose_branch(&exp.slug, &exp.branches, id)?.clone().slug,
             });
         }
     }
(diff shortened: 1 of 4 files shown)


nimbus/src/evaluator.rs
```diff
@@ -48,18 +48,18 @@ impl Bucket {
 pub fn filter_enrolled(id: &Uuid, experiments: &[Experiment]) -> Result<Vec<EnrolledExperiment>> {
     let mut res = Vec::with_capacity(experiments.len());
     for exp in experiments {
-        let bucket_config = exp.arguments.bucket_config.clone();
```
rfk, line 51: The schema has been flattened, there is no longer a separate `arguments` thing.

nimbus/src/http_client.rs
```diff
@@ -199,36 +169,33 @@ mod tests {
...
-                    proposed_duration: Some(28),
-                    proposed_enrollment: 7,
-                    reference_branch: Some("control".to_string()),
-                    features: vec![],
-                    branches: vec![
-                        Branch { slug: "control".to_string(), ratio: 1, feature: FeatureConfig { feature_id: Group::Cfr, enabled: true, value: None } },
-                        Branch { slug: "treatment-variation-b".to_string(), ratio: 1, feature: FeatureConfig { feature_id: Group::Cfr, enabled: true, value: None } }
-                    ]
-                },
+            slug: "mobile-a-a-example".to_string(),
```
rfk, line 175: This matches the example mobile experiment added in https://github.com/mozilla/nimbus-shared/pull/123

piatra:
We don't do client side validation against the schema. This can land.
There's a parallel pull request in https://github.com/mozilla/messaging-system-inflight-assets/pull/157 where we deploy messages from (hand written and linted/tested against the schema) until the Experiments platform can do that for us.

travis79 (approved):
Hooray for flattening the schema!

rfk:
@mhammond flagging this for your context based on conversation in the meeting today.

This PR is up-to-date with the latest schema definitions from https://github.com/mozilla/nimbus-shared, although there are some pending discussions around the details (e.g. whether to [adding an explicit version number](https://github.com/mozilla/nimbus-shared/pull/135)).

The data currently in the "messaging-experiments" bucket on remote settings is *not* compatible with this schema. And I don't believe we have any data deployed that is compatible. There is however a [mobile-a-a.json](https://github.com/mozilla/nimbus-shared/blob/main/data/experiment-recipe-samples/mobile-a-a.json) sample schema that we can use, e.g. by publishing to the dev kinto instance.

We also have a new "nimbus-mobile-experiments" bucket in remote-settings, which IIUC should replace "messaging-experiments" as our default bucket. It's currently empty, but I've asked in slack about how we might get the sample schema published there.

I'm taking this PR out of draft, but it's not clear to me if there's value in merging it until we have an updated sample schema somewhere we can query it. Thoughts?

mhammond:
So IIUC, there's currently no data that conforms to what's currently in our repo, and there never will be. There's also no data that conforms to what we believe is correct (ie, what's done in this PR), but we expect that this will exist at some point in the future. If that's correct, I see no reason not to merge it - it will at least help everyone work out what's wrong with either the future data or the expectations codified here.

travis79:
I agree with Mark. If this is the _current_ schema definition, even if this is the only thing implementing it right now, we can always iterate on this again as the schema evolves.

PR #33, opened 2020-10-01
Introduce nimbus_id

eoger:
Fixes https://jira.mozilla.com/browse/EXP-405.


mhammond (commented):
IIUC the discussion in slack yesterday, we don't really want the app to need to deal with the ID - so it doesn't seem to make sense to force this app to call a new public function, just to pass it back in via `ExperimentConfig` - and not do anything sane if they pass in a different value.

Why not just drop the uid from that struct and manage it entirely ourselves?

travis79:
> IIUC the discussion in slack yesterday, we don't really want the app to need to deal with the ID - so it doesn't seem to make sense to force this app to call a new public function, just to pass it back in via `ExperimentConfig` - and not do anything sane if they pass in a different value.
> 
> Why not just drop the uid from that struct and manage it entirely ourselves?

I'm fine with dropping it from the experiment configuration, but it would be nice to have a way to read the ID across the FFI boundary for recording in telemetry, at least until we can record telemetry directly in Rust. I don't know that this is a hard requirement to start with, but will be needed for DS to be able tie experiments to retention, etc. Hopefully we have the Glean Rust API available by then and we can not have to worry about exposing the ID through the FFI.

nimbus/src/lib.rs
```diff
@@ -89,6 +93,17 @@ impl NimbusClient {
     pub fn update_experiments(&self) -> Result<()> {
         unimplemented!()
     }
+
+    pub(crate) fn nimbus_id(&self) -> Result<Uuid> {
```
travis79, line 97: Until we have the Glean Rust API available, can we make this a `pub` function so that the android-component can access the ID to record in telemetry? I'm happy to have a follow-up issue when we migrate to recording telemetry all in the Rust component to add back the crate-only visibility.

travis79 (changes requested):

travis79 (approved):

PR #37, opened 2020-10-08
IDs refactor

eoger:
Fixes [NIBSSD-66](https://jira.mozilla.com/browse/NIBSSD-66) and [NIBSSD-67](https://jira.mozilla.com/browse/NIBSSD-67).

One of the main pieces of this patch is the introduction of `AvailableRandomizationUnits`, passed by the app at initialization time, which allows experiments to have a custom bucketing strategy.

### DIFF
--- a/nimbus/src/evaluator.rs
+++ b/nimbus/src/evaluator.rs
@@ -9,7 +9,10 @@
 
 //! TODO: Implement the bucketing logic from the nimbus project
 
-use crate::error::{Error, Result};
+use crate::{
+    error::{Error, Result},
+    AvailableRandomizationUnits,
+};
 use crate::{matcher::AppContext, sampling};
 use crate::{Branch, EnrolledExperiment, Experiment};
 use jexl_eval::Evaluator;
@@ -29,7 +32,8 @@ impl Bucket {
 ///
 /// # Arguments:
 ///
-/// - `id` The user's id
+/// - `nimbus_id` The auto-generated nimbus_id
+/// - `available_randomization_units`: The app provded available randomization units
 /// - `experiments` A list of experiments, usually retrieved from the network or persisted storage
 ///
 /// # Returns:
@@ -45,12 +49,29 @@ impl Bucket {
 /// - If the bucket sampling failed (i.e we could not find if the user should or should not be enrolled in the experiment based on the bucketing)
 /// - If an error occurs while determining the branch the user should be enrolled in any of the experiments
 #[allow(dead_code)]
-pub fn filter_enrolled(id: &Uuid, experiments: &[Experiment]) -> Result<Vec<EnrolledExperiment>> {
+pub fn filter_enrolled(
+    nimbus_id: &Uuid,
+    available_randomization_units: &AvailableRandomizationUnits,
+    experiments: &[Experiment],
+) -> Result<Vec<EnrolledExperiment>> {
+    let nimbus_id = nimbus_id.to_string();
     let mut res = Vec::with_capacity(experiments.len());
     for exp in experiments {
         let bucket_config = exp.bucket_config.clone();
+        let id = match available_randomization_units
+            .get_value(&nimbus_id, &bucket_config.randomization_unit)
+        {
+            Some(id) => id,
+            None => {
+                log::info!(
+                    "Could not find a suitable randomization unit for {}. Skipping experiment.",
+                    &exp.slug
+                );
+                continue;
+            }
+        };
         if sampling::bucket_sample(
-            vec![id.to_string(), bucket_config.namespace],
+            vec![id.to_owned(), bucket_config.namespace],
             bucket_config.start,
             bucket_config.count,
             bucket_config.total,
@@ -59,7 +80,7 @@ pub fn filter_enrolled(id: &Uuid, experiments: &[Experiment]) -> Result<Vec<Enro
                 slug: exp.slug.clone(),
                 user_facing_name: exp.user_facing_name.clone(),
                 user_facing_description: exp.user_facing_description.clone(),
-                branch_slug: choose_branch(&exp.slug, &exp.branches, id)?.clone().slug,
+                branch_slug: choose_branch(&exp.slug, &exp.branches, &id)?.clone().slug,
             });
         }
     }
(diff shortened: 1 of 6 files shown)


nimbus/src/evaluator.rs
```diff
@@ -288,26 +313,33 @@ mod tests {
...
+        };
         // 299eed1e-be6d-457d-9e53-da7b1a03f10d uuid fits in start: 0, count: 2000, total: 10000 with the example namespace, to the treatment-variation-b branch
         // Tested against the desktop implementation
         let id = uuid::Uuid::parse_str("299eed1e-be6d-457d-9e53-da7b1a03f10d").unwrap();
-        let enrolled = filter_enrolled(&id, &experiments).unwrap();
+        let enrolled = filter_enrolled(&id, &available_randomization_units, &experiments).unwrap();
         assert_eq!(enrolled.len(), 1);
         assert_eq!(enrolled[0].slug, "TEST_EXP1");
-        // 542213c0-9aef-47eb-bc6b-3b8529736ba2 uuid fits in start: 2000, count: 3000, total: 10000 with the example namespace, to the control branch
-        // Tested against the desktop implementation
```
eoger, line 308: The bucketing here *was not* tested against a reference implementation, do we have any tests that covers `normandy_id` or `client_id` that I could replicate?

nimbus/src/evaluator.rs
```diff
@@ -59,9 +80,13 @@ pub fn filter_enrolled(id: &Uuid, experiments: &[Experiment]) -> Result<Vec<Enro
                 slug: exp.arguments.slug.clone(),
                 user_facing_name: exp.arguments.user_facing_name.clone(),
                 user_facing_description: exp.arguments.user_facing_description.clone(),
-                branch_slug: choose_branch(&exp.arguments.slug, &exp.arguments.branches, id)?
-                    .clone()
-                    .slug,
+                branch_slug: choose_branch(
```
eoger, line 83: IIUC branch bucketing should be solely based on the `nimbus_id`? Am I correct in this assumption?

travis79: I had assumed that it would use the same id as the `filter_enrolled`, which would be either the nimbus_id, or whatever alternate "known" id was selected by the experiment. Maybe we should verify this is still the case with the current desktop/js implementation.

nimbus/src/evaluator.rs
travis79: I don't think we have any test's here that replicate Normandy or Telemetry ids, but maybe something in the [nimbus-shared/test](https://github.com/mozilla/nimbus-shared/tree/main/test) repo covers what you are looking for?

nimbus/src/evaluator.rs
eoger: Sounds like you're right: https://searchfox.org/mozilla-central/source/toolkit/components/messaging-system/experiments/ExperimentManager.jsm#201

k88hudson: That's correct, assuming the `randomizationUnit` in the experiment's `bucketConfig` is `nimbus_id`, you should use value of `nimbus_id`.

nimbus/src/http_client.rs
```diff
@@ -122,7 +122,7 @@ mod tests {
             {
                 "id": "ABOUTWELCOME-PULL-FACTOR-REINFORCEMENT-76-RELEASE",
                 "enabled": true,
-                "targeting": "(firefoxVersion >= 76 && firefoxVersion < 77 && channel == 'release') || (locale == 'en-US' && [userId, \"aboutwelcome-pull-factor-reinforcement-76-release\"]|bucketSample(0, 2000, 10000) && (!('trailhead.firstrun.didSeeAboutWelcome'|preferenceValue) || 'bug-1637316-message-aboutwelcome-pull-factor-reinforcement-76-rel-release-76-77' in activeExperiments))",
+                "targeting": "(firefoxVersion >= 76 && firefoxVersion < 77 && channel == 'release') || (locale == 'en-US' && [normandy_id, \"aboutwelcome-pull-factor-reinforcement-76-release\"]|bucketSample(0, 2000, 10000) && (!('trailhead.firstrun.didSeeAboutWelcome'|preferenceValue) || 'bug-1637316-message-aboutwelcome-pull-factor-reinforcement-76-rel-release-76-77' in activeExperiments))",
```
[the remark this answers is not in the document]
rfk: Ref https://github.com/mozilla/nimbus-sdk/pull/30

nimbus/src/nimbus.idl
```diff
@@ -20,12 +20,16 @@ dictionary EnrolledExperiment {
     string branch_slug;
 };
 
-dictionary ExperimentConfig {
+dictionary Config {
```
rfk, line 23: Since we're renaming things...is there anything in here that's not related to remote-settings? If not, it may be helpful to rename this to `RemoteSettingsConfig` or similar. It would help set caller expectations in the short-term, and may encourage us to factor the code in a way that would work sensibly with a future external remote-settings client in the longer term.

nimbus/src/nimbus.idl
```diff
@@ -20,12 +20,16 @@ dictionary EnrolledExperiment {
...
     string? server_url;
-    string? uuid;
     string? bucket_name;
 };
 
+dictionary AvailableRandomizationUnits {
```
rfk, line 28: A+ name

nimbus/src/nimbus.idl
```diff
@@ -20,12 +20,16 @@ dictionary EnrolledExperiment {
...
+    string? client_id;
+    string? normandy_id;
```
rfk, line 30: I don't expect we will *ever* support normandy id FWIW, it's a legacy thing from desktop.

nimbus/src/nimbus.idl
rfk: (This would also have the benefit of helping disambiguate "bucket" in the remote-settings sense and "bucket" in the experimentation sense, for the "bucket name" field)

eoger:
Ready for final review :)

travis79 (approved):
LGTM!

PR #38, opened 2020-10-09
Add a 'file system' client and update the experiment cmdline app

mhammond:
This PR adds a "file system" client, intended for developers (and later, I suspect it will be used for tests). I ended up splitting the other stuff out into their own PRs.

If the `server_url` is a `file://` URL, the new "file system client" will be used. In this case, we expect the URL to point to a directory, and we will serve up the .json files as experiment. Also includes a directory where we can start to accumulate "useful" experiments and a test.

### DIFF
--- a/nimbus/src/client/fs_client.rs
+++ b/nimbus/src/client/fs_client.rs
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! A SettingsClient that uses the file-system. Used for developer ergonomics
+//! (eg, for testing against experiments which are not deployed anywhere) and
+//! for tests.
+
+use crate::error::Result;
+use crate::Experiment;
+use crate::SettingsClient;
+use std::ffi::OsStr;
+use std::fs::File;
+use std::io::BufReader;
+use std::path::{Path, PathBuf};
+
+pub struct FileSystemClient {
+    path: PathBuf,
+}
+
+impl FileSystemClient {
+    pub fn new<P: AsRef<Path>>(path: P) -> Result<Self> {
+        Ok(Self {
+            path: path.as_ref().into(),
+        })
+    }
+}
+
+impl SettingsClient for FileSystemClient {
+    fn get_experiments_metadata(&self) -> Result<String> {
+        unimplemented!();
+    }
+
+    fn get_experiments(&self) -> Result<Vec<Experiment>> {
+        log::info!("reading experiments in {}", self.path.display());
+        let mut res = Vec::new();
+        // Skip directories and non .json files (eg, READMEs)
+        let json_ext = Some(OsStr::new("json"));
+        let filenames = self
+            .path
+            .read_dir()?
+            .filter_map(Result::ok)
+            .map(|c| c.path())
+            .filter(|f| f.is_file() && f.extension() == json_ext);
+        for child_path in filenames {
+            let file = File::open(child_path.clone())?;
+            let reader = BufReader::new(file);
+            match serde_json::from_reader::<_, Experiment>(reader) {
+                Ok(exp) => res.push(exp),
+                Err(e) => {
+                    log::warn!(
+                        "Malformed experiment found! File {},  Error: {}",
+                        child_path.display(),
+                        e
+                    );
+                }
+            }
+        }
+        Ok(res)
+    }
+}
--- a/nimbus/src/client/mod.rs
+++ b/nimbus/src/client/mod.rs
@@ -1,0 +1,38 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+mod fs_client;
+mod http_client;
+use crate::error::{Error, Result};
+use crate::Experiment;
+use crate::RemoteSettingsConfig;
+use fs_client::FileSystemClient;
+use http_client::Client;
+use url::Url;
+
+pub(crate) fn create_client(
+    config: RemoteSettingsConfig,
+) -> Result<Box<dyn SettingsClient + Send>> {
+    // XXX - double-parsing the URL here if it's not a file:// URL - ideally
+    // config would already be holding a Url and we wouldn't parse here at all.
+    let url = Url::parse(&config.server_url)?;
+    Ok(if url.scheme() == "file" {
+        // Everything in `config` other than the url/path is ignored for the
+        // file-system - we could insist on a sub-directory, but that doesn't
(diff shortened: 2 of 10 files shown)


mhammond:
(oh, new client needs tests too)

nimbus/src/client/fs_client.rs
```diff
@@ -0,0 +1,64 @@
...
+        for child in self.path.read_dir()? {
+            if let Ok(child) = child {
+                let child_path = child.path();
+                if !child_path.is_file() {
+                    log::info!("skipping {} - not a file", child_path.display());
+                }
+                // Only treat files with .json as experiments - we might end up
+                // with README etc?
+                if child_path.extension() != Some(OsStr::new(".json")) {
+                    log::info!("skipping {} - not JSON", child_path.display());
```
mhammond, line 46: doh - needs `continue`!

nimbus/src/client/fs_client.rs
```diff
@@ -0,0 +1,64 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! A SettingssClient that uses the file-system. Used for developer ergonomics
```
rfk, line 5: nit: double "s" in "Settingss"

nimbus/src/client/fs_client.rs
```diff
@@ -0,0 +1,64 @@
...
+                let file = File::open(child_path.clone())?;
+                let reader = BufReader::new(file);
+                match serde_json::from_reader::<_, Experiment>(reader) {
+                    Ok(exp) => res.push(exp),
+                    Err(e) => {
+                        log::warn!(
+                            "Malformed experiment found! File {},  Error: {}",
+                            child_path.display(),
+                            e
+                        );
```
rfk, line 57: For the future, we may want to consider pulling this "validate the experiment JSON and emit logs/metrics about it" logic up to shared code rather than duplicating it; it's fine here for now though.

nimbus/src/client/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+mod fs_client;
+mod http_client;
+use crate::error::{Error, Result};
+use crate::Experiment;
+use fs_client::FileSystemClient;
+use http_client::Client;
+use url::Url;
+
+use crate::ExperimentConfig;
```
rfk, line 13: Naming nit: the name "experiment config" seems pretty misleading, it seems like it should be "remote settings config" or similar.

nimbus/src/client/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+pub(crate) fn create_client(
+    collection_name: &str,
+    config: ExperimentConfig,
+) -> Result<Box<dyn SettingsClient>> {
+    // XXX - double-parsing the URL here if it's not a file:// URL - ideally
+    // config would already be holding a Url and we wouldn't parse here at all.
+    let url = Url::parse(&config.server_url)?;
+    Ok(if url.scheme() == "file" {
+        // collection_name is ignored for the file-system - we could insist
```
rfk, line 23: In fact, it's not obvious to me why `collection_name` is passed as a separate argument rather than as an additional property of the `config`.

rfk (commented):
I like the way this is shaping up. I didn't give it a fully thorough review because it's still in draft, but so far looks good to me.

travis79:
> This will break @travis79's a-c PR as he passes NULL for the config. I can revert this part of it if you like, but I think this is the right thing to do longer term.

Not a problem, it's a one line change in a PR that hasn't landed yet. I agree that the defaults should be owned by the app or at least the A-C component. I'll take a closer look at this first thing tomorrow.

nimbus/src/client/fs_client.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    fn get_experiments(&self) -> Result<Vec<Experiment>> {
+        log::info!("reading experiments in {}", self.path.display());
+        let mut res = Vec::new();
+        for child in self.path.read_dir()? {
+            if let Ok(child) = child {
+                let child_path = child.path();
+                if !child_path.is_file() {
+                    log::info!("skipping {} - not a file", child_path.display());
```
rfk, line 41: `continue` here as well?

Alternately, you could consider factoring this loop as a `filter().map().collect()` iterator chain, where the `filter` step removes these non-JSON items and the `map` step does the parsing of what's left.

I enjoy Rust iterators, but I don't feel strongly about other people using them, so please take or leave that suggestion :-)

nimbus/src/client/mod.rs
rfk: This comment about `collection_name` is out of date now and should be removed.

nimbus/src/lib.rs
```diff
@@ -33,7 +33,7 @@ const DB_KEY_NIMBUS_ID: &str = "nimbus-id";
 /// It should hold all the information needed to communicate a specific user's
 /// experimentation status
 pub struct NimbusClient {
-    http_client: Client,
+    http_client: Box<dyn SettingsClient + Send>,
```
rfk, line 36: naming nit: the `http_client` here might be better called `settings_client` after this refactor, but I'm not sure how invasive a change that would be.

nimbus/tests/test_fs_client.rs
```diff
@@ -0,0 +1,38 @@
...
+    };
+
+    let tmp_dir = TempDir::new("test_fs_client-test_simple")?;
+
+    let aru = AvailableRandomizationUnits { client_id: None };
+    let client = NimbusClient::new(Default::default(), tmp_dir.path(), config, aru)?;
+
+    let experiments = client.get_all_experiments()?;
+    assert_eq!(experiments.len(), 1);
+    assert_eq!(experiments[0].slug, "secure-gold");
```
rfk, line 34: IIUC this will have logged an error about failing to parse the `README` file. Probably not worth it for this PR, but worth thinking about how we might surface those unexpected states to calling code, both for testing purposes and for metrics purposes.

rfk (approved):
I left some nits but overall LGTM 👍

nimbus/src/client/fs_client.rs
mhammond: thanks - yeah, I used an iterator for most of this, except the file loop where I open and parse them.

nimbus/src/lib.rs
mhammond: I changed the `collection_name` client as I think it's still worth calling out it ignored everything in `RemoteSettingsClient` except the URL. Changing to `settings_client` wasn't at all intrusive, so 👍

nimbus/tests/test_fs_client.rs
mhammond: yeah, especially as we don't really expect this to be used by production code. I *hope* it will be useful to developers to avoid waiting for a server deploy before they can start working on an experiment, but that remains to be seen.

travis79 (approved):

PR #39, opened 2020-10-13
A bit of a straw-man for how persistence might work.

mhammond:
Wire up persistence of enrolled experiments.

    * experiments and enrollment status are all stored in the DB. Only
      fetches experiments in the constructor if there are none in the DB,
      which doesn't smell like the right thing long term - otherwise though,
      you must explicitly call `update_experiments()`. We should work out
      what behaviour makes sense for apps here.

    * Fully implements the `opt_out` etc methods. Any experiment can be opted in
      to and opted out of, with this stored in the DB. Calling
      `update_experiments()` will not change the opted-on/out status - there's
      a new 'reset_enrollment` function that does though. Note that opting
      out is fully supported, but the expectation is that this must only be
      used in a dev or test context.

    * When `update_experiments()` is called, enrollments for experiments
      which no longer exist are removed, and new experiments are evaluated
      for enrollment. We don't emit any events yet, but we could. We do log
      so you can see things happening. We don't support experiments being paused,
      nor start and end dates for experiments, etc.

    * It allows you to opt in/out/etc to non-existing experiments - but they will
      be removed when `update_experiments()` is called (ie, it's treated as
      though the experiment was removed. We could fix this, but meh.

    * Out command-line utility `experiment.rs` supports all of this - use --help.

[edited to reflect current commit message]

### DIFF
--- a/nimbus/Cargo.lock
+++ b/nimbus/Cargo.lock
@@ -450,6 +450,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
 
 [[package]]
+name = "fuchsia-cprng"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a06f77d526c1a601b7c4cdd98f54b5eaabffc14d5f2f0296febdc7f357c6d3ba"
+
+[[package]]
 name = "fuchsia-zircon"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -917,7 +923,7 @@ dependencies = [
  "httparse",
  "lazy_static",
  "log",
- "rand",
+ "rand 0.7.3",
  "regex",
  "serde_json",
  "serde_urlencoded",
@@ -975,6 +981,7 @@ dependencies = [
  "serde_derive",
  "serde_json",
  "sha2 0.9.1",
+ "tempdir",
  "thiserror",
  "uniffi",
  "uniffi_build",
@@ -1237,6 +1244,19 @@ dependencies = [
 
 [[package]]
 name = "rand"
+version = "0.4.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "552840b97013b1a26992c11eac34bdd778e464601a4c2054b5f0bff7c6761293"
+dependencies = [
+ "fuchsia-cprng",
+ "libc",
+ "rand_core 0.3.1",
+ "rdrand",
+ "winapi 0.3.9",
+]
+
+[[package]]
+name = "rand"
 version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6a6b1679d49b24bbfe0c803429aa1874472f50d9b363131f0e89fc356b544d03"
@@ -1244,7 +1264,7 @@ dependencies = [
  "getrandom",
  "libc",
  "rand_chacha",
- "rand_core",
+ "rand_core 0.5.1",
  "rand_hc",
 ]
 
@@ -1255,8 +1275,23 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f4c8ed856279c9737206bf725bf36935d8666ead7aa69b52be55af369d193402"
 dependencies = [
  "ppv-lite86",
- "rand_core",
+ "rand_core 0.5.1",
 ]
+
+[[package]]
+name = "rand_core"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7a6fdeb83b075e8266dcc8762c22776f6877a63111121f5f8c7411e5be7eed4b"
+dependencies = [
+ "rand_core 0.4.2",
+]
+
+[[package]]
+name = "rand_core"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9c33a3c44ca05fa6f1807d8e6743f3824e8509beca625669633be0acbdf509dc"
 
 [[package]]
 name = "rand_core"
@@ -1273,7 +1308,16 @@ version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ca3129af7b92a17112d59ad498c6f81eaf463253766b90396d39ea7a39d6613c"
 dependencies = [
- "rand_core",
+ "rand_core 0.5.1",
+]
+
+[[package]]
+name = "rdrand"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "678054eb77286b51581ba43620cc911abf02758c91f93f479767aed0f90458b2"
+dependencies = [
+ "rand_core 0.3.1",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


k88hudson:
Just a general note – allowing folks to knowingly *opt in* introduces bias that we would consider unacceptable for randomized controlled experiments (see point 2 in the (Bucketing Statistical Requirements](https://docs.google.com/document/d/1jZQ--ozuQ0_Su-lh7-dX-3yqmWsIan2QPYyTt6lv4ow/edit#heading=h.kd07kyqteh85).

There are real use cases for opt-in of course but the data requirements are different – I'd recommend not handling this for now (or kicking users out of experiment data, which we can talk through if you want)

travis79:
> Just a general note – allowing folks to knowingly _opt in_ introduces bias that we would consider unacceptable for randomized controlled experiments (see point 2 in the (Bucketing Statistical Requirements]([docs.google.com/document/d/1jZQ--ozuQ0_Su-lh7-dX-3yqmWsIan2QPYyTt6lv4ow/edit#heading=h.kd07kyqteh85](https://docs.google.com/document/d/1jZQ--ozuQ0_Su-lh7-dX-3yqmWsIan2QPYyTt6lv4ow/edit#heading=h.kd07kyqteh85)).
> 
> There are real use cases for opt-in of course but the data requirements are different – I'd recommend not handling this for now (or kicking users out of experiment data, which we can talk through if you want)

Having an opt-in for testing purposes seems reasonable, but I agree that allowing anyone to opt-in to an experiment may not line up with specific device or targeting requirements for the experiment or create bias. But, just because we have the ability to opt in at the nimbus-sdk level doesn't necessarily mean that an application has to expose this functionality to the user.

k88hudson:
> just because we have the ability to opt in at the nimbus-sdk level doesn't necessarily mean that an application has to expose this functionality to the user.

True, my concerns are all around exposing it to users so if you've got another use case (the testing one makes sense) go for it!

nimbus/src/evaluator.rs
```diff
@@ -28,65 +29,65 @@ impl Bucket {
...
-                branch_slug: choose_branch(&exp.slug, &exp.branches, &id)?.clone().slug,
+                status: EnrollmentStatus::Error {
+                    reason: "No randomization unit".into(),
+                },
             });
         }
-    }
-    Ok(res)
+    };
+    let status = if sampling::bucket_sample(
```
travis79, line 74: We are also not checking any filtering, but we should probably check that the `application` field in the experiment matches the `app_id` field in the `AppContext` at some point in here, in fact, it might be best to check that first to avoid the enrollment calculations if this isn't the intended app. Although, I suppose this isn't _really_ important until we have more than one application using nimbus on mobile...

nimbus/src/evaluator.rs
```diff
@@ -28,65 +29,65 @@ impl Bucket {
...
+/// determine if the user is actually enrolled.
 /// # Errors:
 ///
 /// The function can return errors in one of the following cases (but not limited to):
 ///
 /// - If the bucket sampling failed (i.e we could not find if the user should or should not be enrolled in the experiment based on the bucketing)
 /// - If an error occurs while determining the branch the user should be enrolled in any of the experiments
-#[allow(dead_code)]
-pub fn filter_enrolled(
+pub fn evaluate_enrollment(
```
travis79, line 48: Huge +1 from me on the name change here.

nimbus/src/lib.rs
```diff
@@ -51,68 +47,102 @@ impl NimbusClient {
...
-                db.put(DB_KEY_NIMBUS_ID, &nimbus_id)?;
+                db.put(StoreId::Meta, DB_KEY_NIMBUS_ID, &nimbus_id)?;
                 nimbus_id
             }
         })
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct EnrolledExperiment {
```
travis79, line 139: Can we add something with the enrollment status to this, at least until we can record to Glean in Rust? That way the A-C component can record the enrollment status in telemetry.

travis79 (commented):
I'm definitely onboard with this approach!

nimbus/src/evaluator.rs
mhammond: I wired up the existing `targeting` function, but didn't check the app field yet - we have a `struct Matcher` that, according to its comments, implies the server is going to send one of them in an experiment and we should be matching that way - but that contradicts with the server. So I think we need a new ticket on file to wire up "application matching" - and it that turns out to be literally just a string comparison, that's fine and we can remove the matcher.

nimbus/src/lib.rs
mhammond: We can - but I'm not quite sure what you are asking for. The only reasons you can be enrolled are because you were selected, or you opted-in. Opting-in is a test only feature, so it's not clear it makes sense.

So maybe what you are really after is a list of experiments you were targeted for, but were not enrolled in? Or we rename this struct to something else and make some of the elements `Option<>` so it can hold both enrolled and "not selected" experiments? Let's chat more about this tomorrow.

mhammond:
I think this is ready to roll - still a fair bit to be done, but everyone seems to think this is heading in the right direction, so 🚀

nimbus/Cargo.lock
```diff
@@ -1235,6 +1242,19 @@ dependencies = [
  "proc-macro2",
 ]
 
+[[package]]
```
eoger, line 1245: I assume this is fine since we don't need high-quality random numbers, but I'd be nice if we could make `rc_crypto` available to other rust projects :/

nimbus/src/enrollment.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+    // and existing enrollments.
+    let enrollments = db.collect_all::<ExperimentEnrollment>(StoreId::Enrollments)?;
+    let mut map_enrollments = HashMap::with_capacity(enrollments.len());
+    for e in enrollments {
+        all_slugs.insert(e.slug.clone());
+        map_enrollments.insert(e.slug.clone(), e);
+    }
+
+    // XXX - we want to emit events for many of these things, but we don't have
```
eoger, line 105: We could also return a "Diff" structure to the Kotlin layer, and let them emit observable events to the app.

nimbus/src/lib.rs
```diff
@@ -51,68 +47,106 @@ impl NimbusClient {
...
+            available_randomization_units,
             app_context,
             db,
             nimbus_id,
-        })
+        };
+        // This is a little suspect but it's not clear what the right thing is.
+        // If we have no experiments we assume this is a first run so fetch
+        // them now.
+        if instance.get_all_experiments()?.is_empty() {
```
eoger, line 63: We're gonna have to move this so we don't do I/O in the constructor, I wonder if it'll make it harder?

eoger (approved):
LGTM, let's not let this great piece of code rot and improve upon it instead.

nimbus/src/lib.rs
mhammond: We already do network and disk io in the constructor (disk io is opening the db). We do need to think about that, but I don't think this patch makes things any worse really.

nimbus/Cargo.lock
mhammond: I think this dep is just for `temp_dir` - that's the only new package I added.

travis79 (approved):
This is a huge improvement! No complaints from me.

nimbus/src/lib.rs
dmose: Because we're planning to get Nimbus vendored into mozilla-central before we do the work of sorting out the details of the IO threading stuff, I/O in the constructor will prevent us from landing in mozilla-central.  Which is to say, we need a very simple xcpshell test to land, and we've got one: it tries to construct an object, and right now that causes viaduct on desktop to assert and crash because it ends up trying to do IO on the main thread.  I don't know that this needs to necessarily block this patch from landing, but I do want to raise a flag here that this sort of I/O will start to block desktop vendoring soon.  CC @piatra

mhammond: Ok, I tweaked this to avoid the network IO in the ctor (but note that disk IO remains - we open the DB). Note also  that #38 might provide better general test strategy for desktop - you could configure the sdk with a file:// URL and we'd never do network IO - it's probably either that, or a test http server which presumably would *still* need to be backed by the file system, so we might as well cut out the middle-man :)

PR #40, opened 2020-10-15
Modify Cargo.toml and friends to use patched version of jexl

dmose:
This bumps jexl-rs version to 0.1.7, which contains @eoger's checkin of the parsers.  That shrinks the number of Nimbus dependencies from 152 to 199, in addition to getting rid of some truly giant files!


eoger (approved):
LGTM!

PR #41, opened 2020-10-19
Make RemoteSettingsConfig and the server URL inside it non-optional

mhammond:
I split this from #38 mainly for ease of review. Note that @rfk and I discussed adding constants for the default server URLs, but given the discussion started by @jhugman it's not clear yet that is actually desirable - string constants in the consuming code are probably less likely to be overlooked than, eg, `Servers.RELEASE`

Explicitly flagging @travis79 for review as it will (somewhat trivially) break his a-c PR.

### DIFF
--- a/nimbus/src/config.rs
+++ b/nimbus/src/config.rs
@@ -14,6 +14,7 @@
 /// - `bucket_name`: The name of the bucket containing the collection on the server
 #[derive(Debug, Clone)]
 pub struct RemoteSettingsConfig {
-    pub server_url: Option<String>,
-    pub bucket_name: Option<String>,
+    pub server_url: String,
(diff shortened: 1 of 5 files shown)


nimbus/src/config.rs
```diff
@@ -14,6 +14,6 @@
 /// - `bucket_name`: The name of the bucket containing the collection on the server
 #[derive(Debug, Clone)]
 pub struct RemoteSettingsConfig {
-    pub server_url: Option<String>,
+    pub server_url: String,
```
travis79, line 17: While we are at it, is there any reason not to include the `collection_name` within this struct?

travis79 (approved):
Looks good to me!

PR #42, opened 2020-10-19
Avoid I/O in constructor

eoger:
Fixes https://jira.mozilla.com/browse/NIBSSD-71.
This change makes the DB initialize lazily.

### DIFF
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -14,16 +14,17 @@ mod sampling;
 #[cfg(debug_assertions)]
 pub use evaluator::evaluate_enrollment;
 
-use ::uuid::Uuid;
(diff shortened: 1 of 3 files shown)


nimbus/src/lib.rs
```diff
@@ -35,27 +36,27 @@ pub struct NimbusClient {
     http_client: Client,
     available_randomization_units: AvailableRandomizationUnits,
     app_context: AppContext,
-    db: Database,
-    nimbus_id: Uuid,
```
eoger, line 39: `nimbus_id` should be cached by LLDB anyway, so I dropped our own cache of it in favor of simplicity.

mhammond (approved):

PR #44, opened 2020-10-20
Add some enhancements to the 'experiment' example

mhammond:
* Reinstates the `gen-uuid` command, and optionally sets
  the discovered uuid in the database. Also shows progress
  so you can see if you are in a loop that's likely to never
  complete.

* Adds a `brute-force` command which shows how any number of enrollments
  actually ends up being distributed.

### DIFF
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -142,6 +142,13 @@ impl NimbusClient {
         })
     }
 
+    // Sets the nimbus ID - TEST ONLY - should not be exposed to real clients.
+    // (Useful for testing so you can have some control over what experiments
+    // are enrolled)
+    pub fn set_nimbus_id(&self, uuid: &Uuid) -> Result<()> {
+        self.db()?.put(StoreId::Meta, DB_KEY_NIMBUS_ID, uuid)
+    }
+
     fn db(&self) -> Result<&Database> {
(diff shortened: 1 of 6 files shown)


nimbus/src/lib.rs
```diff
@@ -142,6 +142,13 @@ impl NimbusClient {
         })
     }
 
+    // Sets the nimbus ID - TEST ONLY - should not be exposed to real clients.
+    // (Useful for testing so you can have some control over what experiments
+    // are enrolled)
+    pub fn set_nimbus_id(&self, uuid: &Uuid) -> Result<()> {
+        self.db()?.put(StoreId::Meta, DB_KEY_NIMBUS_ID, uuid)
+    }
```
travis79, line 150: This "for testing" functionality could potentially be accomplished without an explicit `set_nimbus_id` function if you passed in a `client_id` into the init and set up the experiment configuration to use it instead of the `nimbus_id`

mhammond: I don't quite see how that would work - which ID to use is part of the experiment definition, and the example can't influence that if it's using a server.

(Another vague idea I had is that this experiment could work directly on the database used by an Android emulator - eg, you could use this app to ensure the ID was one which was enrolled in a specific experiment, but I have no idea how useful that would be in practice)

travis79 (approved):

PR #45, opened 2020-10-21
Work around geckojs build bustage from UniFFI; fixes JIRA:SDK-87

dmose:


eoger (approved):
It's hacky, but the issue shouldn't block us.

dmose:
The original patch didn't build everywhere, because the changes reflected in the IDL weren't propagated through to the hand-coded stuff in the various bindings.  After chatting with @mhammond, it seemed to make more sense to simply make client_id not be optional, since that's presumably where it wants to end up anyway.  Sorry if this is the wrong way forward; wanted to unblock folks working on gecko...

dmose:
Bah, not compiling on Circle.  I'll have to look more to see why it's building locally for me but not here..

nimbus/src/evaluator.rs
```diff
@@ -71,7 +71,7 @@ pub fn evaluate_enrollment(
         match available_randomization_units
```
mhammond, line 71: We no longer need this match - ie, this line can read `let id = available_randomization_units.get_value(...);`

mhammond (changes requested):
LGTM, and I agree that longer term there's no need to support consumers who can't supply a `client_id` - but we should clean up that one block.

nimbus/src/evaluator.rs
dmose: Fixed.

mhammond (approved):

PR #47, opened 2020-10-23
Make AvailableRandomizationUnits.client_id optional and add a dummy field, reverting #45

mhammond:
How does this look? I don't think there's anything we can do here to avoid JS code dealing with the dummy field, although I think @travis's a-c PR can hide that from Android app consumers.

(I'm also assuming an `i8` works around the problem and is lighter-weight than a string - @dmose is it your understanding that should work OK?)

### DIFF
--- a/nimbus/examples/experiment.rs
+++ b/nimbus/examples/experiment.rs
@@ -205,9 +205,7 @@ fn main() -> Result<()> {
         bucket_name: bucket_name.to_string(),
     };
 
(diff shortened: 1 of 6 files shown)


eoger (approved):

dmose:
I don't see why type would matter.  Thanks for putting this together!  r=dmose

dmose (approved):

PR #48, opened 2020-10-26
Bump uniffi dependency to 0.4.0 to fix license goofiness for m-c

dmose:
Obviously, CI won't pass and I won't try to land this until after the uniffi PR lands.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,9 +10,6 @@ jobs:
           name: Version information
           command: rustc --version; cargo --version; rustup --version
       - run:
-          name: Install uniffi
-          command: cargo install --version 0.3.0 uniffi_bindgen
-      - run:
           name: Calculate dependencies
           command: cd nimbus && cargo generate-lockfile
       - restore_cache:
--- a/nimbus/Cargo.lock
+++ b/nimbus/Cargo.lock
@@ -1515,9 +1515,9 @@ checksum = "f7fe0bb3479651439c9112f72b6c505038574c9fbb575ed1bf3b797fa39dd564"
 
 [[package]]
 name = "uniffi"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "421f78982027ce32235052fba9d147c7094596a6404b71cb761ca653d53786b2"
+checksum = "61c4ab58b81243770e57108ea885c938ea81c6fcb89d46a96f6a33842cf6f67a"
 dependencies = [
  "anyhow",
  "bytes",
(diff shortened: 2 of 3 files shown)


rfk (approved):
uniffi v0.4.0 is out, and this LGTM 👍

dmose:
I saw exactly the same error locally.  It turned that I had `uniffi_bindgen 0.3.0` installed in `~/.cargo/bin`.  What I had to do to fix it was do `cargo uninstall uniffi_bindgen && cargo install uniffi_bindgen`.  @rfk, does mean that we will need to update the docker image shortly before or after we land this PR?

```
   Compiling nimbus-sdk v1.0.0 (/root/project/nimbus)
error: failed to run custom build command for `nimbus-sdk v1.0.0 (/root/project/nimbus)`

Caused by:
  process didn't exit successfully: `/root/project/nimbus/target/debug/build/nimbus-sdk-2e4b887a0f7f59eb/build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=./src/nimbus.idl

  --- stderr
  Error: The `uniffi` dependency version (0.4.0) is different than `uniffi-bindgen` own version (0.3.0). Please rectify the problem and try again.
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error while generating scaffolding code', build.rs:6:60
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

rfk:
@dmose aha, looks like `.circleci/config.yml` is hard-coding a specific version of uniffi to install for the tests, we should either update it there as well, or figure out how to have that install the version specified in `Cargo.tom`.

mhammond:
> @dmose aha, looks like `.circleci/config.yml` is hard-coding a specific version of uniffi to install for the test

The megazord branch has a similar issue - [here's one place](https://github.com/mozilla/application-services/blob/nimbus-megazord/megazords/full/android/build.gradle#L67) and I think there are others.

Ironically, uniffi-build says:

    // Calling the command line helps making sure that the generated swift/Kotlin/whatever
    // bindings were generated with the same version of uniffi as the Rust scaffolding code.

but as we see, that doesn't seem quite true. Is this still the right tradeoff? Regardless, this seems likely to bite us more and more

dmose:
So it sounds like the proposal is to switch to _always_ using
builtin_bindgen, which seems reasonable to me.  I'd love to hear @eoger's
thoughts on this...

eoger:
I'm against using `builtin_bindgen`, because it can introduce subtle breakage between Kotlin and Rust layers, but I won't block it.

dmose:
@eoger Can you unpack that subtle breakage a bit more?  Both so that we as developers could recognize it in the future, but also to see if we can brainstorm a fix that doesn't have that issue, while also not having the fragility of the current setup?

I mean, I could image a world where the right thing to do is simpy have a build_dependency on `uniffi_bindgen` instead of on `uniffi_build`, and clients would be expect to write their own logic to generate artifacts from udl files.  Does that sounds like a better option?

dmose:
Couldn't one even make the argument that uniffi_bindgen should be available as a feature of uniffi itself,  I can't imagine ever wanting to use different versions of those things?

rfk:
It's also worth noting that we have some additional layers of protection against dylib/bindings incompatibility now, such as https://github.com/mozilla/uniffi-rs/issues/333 (which includes the uniffi version as part of the fingerprint to include in the symbol names).

dmose:
As discussed in today's meeting, not blocking on finding the ideal solution.  Fixes JIRA:SDK-93

PR #49, opened 2020-10-27
Add schema version validation

travis79:
This is still a WIP for adding schema version validation.

Design requirements:
* SDK maintains awareness of the supported schema version
* Check the "schemaVersion" field before doing any other processing or validation of experiment config.
* If the major version number is higher than the version supported by the SDK, ignore the experiment (in future we would probably emit some diagnostic telemetry here)
* Document how this is meant to work for clients and their responsibility in testing and updating the nimbus-sdk.

Open question:
* Is backwards compatibility for all previous versions assumed? If not - what are the limits?

### DIFF
--- a/nimbus/src/client/http_client.rs
+++ b/nimbus/src/client/http_client.rs
@@ -13,10 +13,11 @@
 //!
 //! But the simple subset implemented here meets our needs for now.
 
+use std::str::FromStr;
+
 use crate::config::RemoteSettingsConfig;
 use crate::error::{Error, Result};
-use crate::Experiment;
-use crate::SettingsClient;
+use crate::{Experiment, SettingsClient, SCHEMA_VERSION};
 use url::Url;
 use viaduct::{status_codes, Request, Response};
 
@@ -66,6 +67,27 @@ impl SettingsClient for Client {
         let data = resp.get("data").ok_or(Error::InvalidExperimentResponse)?;
         let mut res = Vec::new();
         for exp in data.as_array().ok_or(Error::InvalidExperimentResponse)? {
+            // Validate the schema major version matches the supported version
+            let exp_schema_version = match exp.get("schemaVersion") {
+                Some(ver) => {
+                    serde_json::from_value::<String>(ver.to_owned()).unwrap_or("".to_string())
+                }
+                None => {
+                    log::trace!("Missing schemaVersion: {:#?}", exp);
+                    continue;
+                }
+            };
+            let schema_maj_version = exp_schema_version.split(".").next().unwrap_or("");
+            let schema_version: i32 = FromStr::from_str(schema_maj_version).unwrap();
+            if schema_version != SCHEMA_VERSION {
+                log::warn!(
+                    "Schema version mismatch: Expected version {}, discarding experiment with version {}",
+                    SCHEMA_VERSION, schema_version
+                );
+                // Schema version mismatch
+                continue;
+            }
+
             match serde_json::from_value::<Experiment>(exp.clone()) {
                 Ok(exp) => res.push(exp),
                 Err(e) => {
@@ -91,42 +113,75 @@ mod tests {
     #[test]
     fn test_get_experiments_from_schema() {
         viaduct_reqwest::use_reqwest_backend();
-        let body = r#"
-        { "data": [
-            {
+        // There are two experiments defined here, one has a "newer" schema version
+        // in order to test filtering of unsupported schema versions.
+        let body = format!(
+            r#"
+        {{ "data": [
+            {{
+                "schemaVersion": "{}.0.0",
                 "slug": "mobile-a-a-example",
                 "application": "reference-browser",
                 "userFacingName": "Mobile A/A Example",
                 "userFacingDescription": "An A/A Test to validate the Rust SDK",
                 "isEnrollmentPaused": false,
-                "bucketConfig": {
+                "bucketConfig": {{
                     "randomizationUnit": "nimbus_id",
                     "namespace": "mobile-a-a-example",
                     "start": 0,
                     "count": 5000,
                     "total": 10000
-                },
+                }},
                 "startDate": null,
                 "endDate": null,
                 "proposedEnrollment": 7,
                 "referenceBranch": "control",
                 "probeSets": [],
                 "branches": [
-                    {
+                    {{
                     "slug": "control",
                     "ratio": 1
-                    },
-                    {
+                    }},
+                    {{
                     "slug": "treatment-variation-b",
                     "ratio": 1
-                    }
+                    }}
                 ]
-            },
-            {
-                "hello": "bye"
-            }
-        ]}
-          "#;
+            }},
+            {{
+                "schemaVersion": "{}.0.0",
+                "slug": "mobile-a-a-example",
+                "application": "reference-browser",
+                "userFacingName": "Mobile A/A Example",
+                "userFacingDescription": "An A/A Test to validate the Rust SDK",
+                "isEnrollmentPaused": false,
+                "bucketConfig": {{
+                    "randomizationUnit": "nimbus_id",
+                    "namespace": "mobile-a-a-example",
+                    "start": 0,
+                    "count": 5000,
+                    "total": 10000
+                }},
+                "startDate": null,
+                "endDate": null,
+                "proposedEnrollment": 7,
+                "referenceBranch": "control",
+                "probeSets": [],
+                "branches": [
+                    {{
+                    "slug": "control",
+                    "ratio": 1
+                    }},
+                    {{
+                    "slug": "treatment-variation-b",
+                    "ratio": 1
+                    }}
+                ]
+            }}
+        ]}}"#,
+            SCHEMA_VERSION,
+            SCHEMA_VERSION + 1
+        );
         let m = mock(
             "GET",
             "/buckets/main/collections/messaging-experiments/records",
@@ -148,6 +203,7 @@ mod tests {
         assert_eq!(
             exp.clone(),
             Experiment {
+                schema_version: "1.0.0".to_string(),
                 slug: "mobile-a-a-example".to_string(),
                 application: "reference-browser".to_string(),
                 user_facing_name: "Mobile A/A Example".to_string(),
--- a/nimbus/src/client/http_client.rs
+++ b/nimbus/src/client/http_client.rs
@@ -13,8 +13,6 @@
 //!
 //! But the simple subset implemented here meets our needs for now.
 
-use std::str::FromStr;
-
 use crate::config::RemoteSettingsConfig;
 use crate::error::{Error, Result};
 use crate::{Experiment, SettingsClient, SCHEMA_VERSION};
@@ -78,9 +76,11 @@ impl SettingsClient for Client {
                 }
             };
             let schema_maj_version = exp_schema_version.split(".").next().unwrap_or("");
-            let schema_version: i32 = FromStr::from_str(schema_maj_version).unwrap();
+            // While "0" is a valid schema version, we have already passed that so reserving zero as
+            // a special value here in order to avoid a panic, and just ignore the experiment.
+            let schema_version: u32 = schema_maj_version.parse().unwrap_or(0);
             if schema_version != SCHEMA_VERSION {
-                log::warn!(
+                log::info!(
                     "Schema version mismatch: Expected version {}, discarding experiment with version {}",
                     SCHEMA_VERSION, schema_version
                 );
@@ -119,7 +119,7 @@ mod tests {
             r#"
         {{ "data": [
             {{
-                "schemaVersion": "{}.0.0",
+                "schemaVersion": "{current_version}.0.0",
                 "slug": "mobile-a-a-example",
                 "application": "reference-browser",
                 "userFacingName": "Mobile A/A Example",
@@ -149,7 +149,7 @@ mod tests {
                 ]
             }},
             {{
-                "schemaVersion": "{}.0.0",
+                "schemaVersion": "{newer_version}.0.0",
                 "slug": "mobile-a-a-example",
                 "application": "reference-browser",
                 "userFacingName": "Mobile A/A Example",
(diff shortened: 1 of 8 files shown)


nimbus/src/lib.rs
```diff
@@ -162,9 +165,22 @@ pub struct EnrolledExperiment {
...
+pub const SCHEMA_VERSION: Version = Version {
+    major: 1,
+    minor: 0,
+    patch: 0,
+    pre: vec![],
+    build: vec![],
+};
+// XXX: In the future it would be nice if this lived in its own versioned crate so that
+// the schema could be decoupled from the sdk so that it can be iterated on while the
+// sdk depends on a particular version of the schema through the cargo.toml.
```
mhammond, line 178: I guess it depends on how the versions are going to be used in practice. I'd expect that a version bump would require corresponding changes to the rust code in the SDK and this version string ensures they are in sync?

mhammond (commented):
This seems fine to me, but I'm not marking it as approved as it's still a draft, so I assume you expect further changes.

travis79:
> This seems fine to me, but I'm not marking it as approved as it's still a draft, so I assume you expect further changes.

Thanks @mhammond, I wanted to get a first pass review of this to make sure I wasn't going off the rails.

I'm talking with Kate about what the backwards compatibility for the schema looks like, so I expect there might be some additions once that is sorted out.

nimbus/src/lib.rs
travis79: Yes, my thought was that this represents the version of the schema that is currently implemented here.  My biggest concern now is with backwards compatibility...  will we need multiple versions of the structs that represent the schema in order to support 2 versions backwards (per the requirement I learned from Kate), and how does that affect our persistent storage, etc.  

I'm in the process of writing up a short doc to define my understanding of what is being asked for with backwards compatibility, so I hope to be able to answer some of the questions I have.

travis79:
This really stretched my Rust skills, please let me know if there are better ways to do this.

travis79:
I also updated the experiment "secure-gold" on the staging server, adding the `schemaVersion` to it.

nimbus/src/lib.rs
```diff
@@ -162,9 +194,20 @@ pub struct EnrolledExperiment {
     pub branch_slug: String,
 }
 
+// XXX: In the future it would be nice if this lived in its own versioned crate so that
+// the schema could be decoupled from the sdk so that it can be iterated on while the
+// sdk depends on a particular version of the schema through the cargo.toml.
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq)]
+#[serde(untagged)]
```
rfk, line 202: I'm concerned about the use of `untagged` here making the system fragile and harder to reason about.

For example, suppose that we define a "2.0.0" schema version. Naively, I'd expect us to want to add a `V2(V2_0_0)` variant to this enum. But what if the JSON data in `V2_0_0` also happens to be parseable as a valid `V1_0_0` struct (perhaps because it adds new fields that the `V1_0_0` parser will ignore)? Whether we end up with a `V1` or a `V2` will depend on which order the variants are declared in the enum, which is non-obvious.

Alternately, we could evolve the `V1_0_0` struct in such a way that it can represent both "1.0.0" and "2.0.0" experiment data, and inspect the `schema_version` attribute at runtime to decide which semantics to impose. I think this would make the consuming code harder to reason about because you might naturally read `Experiment::V1` and assume that it's a "1.x.y" experiment.

It might be cleaner to instead bypass serde's magic here and implement our own switching on the `schema_version` field, something like:

```
pub fn parse_experiment_json(data: &serde_json::Value) -> Result<Experiment> {
  let schema_version = data.get("schema_version").ok_or(Error:InvalidExperimentResponse)?;
  if V1_SEMVER_MATCH(schema_version) {
    return Experiment::V1(serde_json::from_value::<V1_0_0>(data)?)
  }
  if V2_SEMVER_MATCH(schema_version) {
    return Experiment::V2(serde_json::from_value::<V2_0_0>(data)?)
  }
  return Err(Error:UnsupportedSchemaVersion)
}
```

It's a little more boilerplate but I think it would make it more obvious that we're getting the intended result.

nimbus/src/lib.rs
```diff
@@ -154,6 +167,25 @@ impl NimbusClient {
     }
 }
 
+pub trait SchemaRequirement {
+    fn schema_requirement(&self) -> VersionReq;
+}
```
rfk, line 172: I'm not sure this needs to be a trait, as you don't use it as a trait bound anywhere. Would it work if you just `impl NimbusClient` rather than `impl SchemaRequirement for NimbusClient` below?

nimbus/src/evaluator.rs
```diff
@@ -362,10 +397,70 @@ mod tests {
...
+            ],
+            reference_branch: Some("control".to_string()),
+            ..Default::default()
+        });
+
+        // Create a second experiment with a non-matching schema_version
+        let experiment2 = match experiment1.clone() {
+            Experiment::V1(mut e) => {
+                e.schema_version = "2.0.0".to_string();
+                Experiment::V1(e)
```
rfk, line 433: This test is a neat example of where my mental model is getting tripped up - something about creating an `Experiment::V1` struct that has `schema_version: "2.0.0"` seems weird to me, and like the checking is happening in the wrong place. I would expect us to have produced an `Experiment::V2` struct for something with `schema_version: "2.0.0"`, or to have ignored it completely.

(That's not to say that we shouldn't do it this way! Just noting that I found it confusing, and may be worth some additional docs or context-setting or something).

nimbus/src/enrollment.rs
```diff
@@ -124,7 +133,7 @@ pub fn update_enrollments(
                 log::debug!("Experiment '{}' already has enrollment {:?}", slug, enr)
             }
             (Some(exp), None) => {
-                let enr = evaluate_enrollment(nimbus_id, aru, app_context, &exp)?;
+                let enr = evaluate_enrollment(nimbus_id, aru, app_context, &exp, schema_version)?;
```
rfk, line 136: If `evaluate_enrollment` returns an error, IIUC we're going to immediately bail out and return an error from `update_enrollments`, which will in turn fail the call to `NimbusClient.update_experiments`. That might cause us to not be enrolled in other, perfectly valid experiments because we bailed out of the loop early.

I know `evaluate_enrollment` was already fallible so we kind of already had this problem, but I think the introduction of the new `SchemaError` could make it much more likely to happen in practice. Encountering an experiment with an unexpected schema doesn't feel like it should be a fatal error.

rfk (commented):
High-level, I'm not sure I fully understand how this fits into the overall architecture here. I think naively I was expecting the bulk of the schema-version-checking to happen down in the layer where we fetch experiments from the server, but here we're doing it when evaluating enrollment. I left some notes and questions below to help me understand and adjust my mental model.

I think the best way to get to a good shared understanding of how this should work would be to add some higher-level testcases. For example, testing that if we read a "2.0.0" experiment and then a "1.0.0" experiment from the server, we correctly ignore the first and enroll ourselves in the second.

nimbus/Cargo.lock
```diff
@@ -831,6 +831,7 @@ dependencies = [
  "mockito",
  "once_cell",
  "rkv",
+ "semver 0.11.0",
```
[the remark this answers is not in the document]
rfk: Ref https://github.com/mozilla/uniffi-rs/pull/341 for the update in uniffi

nimbus/src/enrollment.rs
```diff
@@ -298,7 +318,7 @@ mod tests {
         // Opt in to a specific branch.
         opt_in_with_branch(&db, "secure-gold", "treatment")?;
         let enrollments = get_enrollments(&db)?;
-        assert_eq!(enrollments.len(), 1);
+        // assert_eq!(enrollments.len(), 1);
```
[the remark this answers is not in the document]
travis79: Oops, yeah I think I had this commented out while debugging a test failure.

nimbus/src/evaluator.rs
```diff
@@ -1,13 +1,14 @@
...
  * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
 
-//! This might be where the bucketing logic can go
-//! It would be different from current experimentation tools
-//! There is a namespacing concept to allow users to be in multiple
-//! unrelated experiments at the same time.
+// ! This might be where the bucketing logic can go
```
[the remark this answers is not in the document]
travis79: I think that this was done to me by a `cargo fmt`, definitely not intentional.  The comments do seem stale so I think it would probably be better to just remove them.

nimbus/Cargo.toml
```diff
@@ -30,6 +30,7 @@ sha2 = "0.9"
 hex = "0.4"
 once_cell = "1"
 uniffi = "0.4"
+semver = "0.11.0"
```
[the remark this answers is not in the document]
travis79: Since we are only really looking at supporting 1-2 versions of the schema at a time, and only really need to look at Major versions, I can probably pare this back down to doing a little string slicing.  The convenience of the crate is good, but I'm not sure it's worth that big of a hit.

nimbus/examples/experiment.rs
```diff
@@ -216,10 +217,11 @@ fn main() -> Result<()> {
...
             println!("======================================");
             println!("Printing all experiments (regardless of enrollment)");
-            nimbus_client
-                .get_all_experiments()?
-                .iter()
-                .for_each(|e| println!("Experiment: {}", e.slug));
+            nimbus_client.get_all_experiments()?.iter().for_each(|e| {
+                match e {
+                    Experiment::V1(e) => println!("Experiment: {}", e.slug),
+                };
```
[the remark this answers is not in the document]
travis79: A lot of this may be just my lack of experience with Rust and knowing what good approaches might be here. Your suggestions sound a lot simpler than having to match enum variants everywhere and keeps the code that matters in fewer places.  Let me take a shot at refactoring this to be more in line with what you are suggesting.

nimbus/src/evaluator.rs
```diff
@@ -50,8 +52,33 @@ pub fn evaluate_enrollment(
     available_randomization_units: &AvailableRandomizationUnits,
     app_context: &AppContext,
     exp: &Experiment,
+    schema_requirement: &VersionReq,
```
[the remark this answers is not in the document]
travis79: My thinking was that passing it as an argument here would make it easier to write a unit test for it.

travis79:
@rfk I updated this to do the schema version checking when it receives the data rather than when it evaluates it.  Ready for another look.

nimbus/src/client/http_client.rs
```diff
@@ -13,10 +13,11 @@
 //!
 //! But the simple subset implemented here meets our needs for now.
 
+use std::str::FromStr;
```
rfk, line 16: If you use `.parse` as I suggest in a separate comment, I don't think you'll need this import.

nimbus/src/client/http_client.rs
```diff
@@ -66,6 +67,27 @@ impl SettingsClient for Client {
...
+                }
+                None => {
+                    log::trace!("Missing schemaVersion: {:#?}", exp);
+                    continue;
+                }
+            };
+            let schema_maj_version = exp_schema_version.split(".").next().unwrap_or("");
+            let schema_version: i32 = FromStr::from_str(schema_maj_version).unwrap();
+            if schema_version != SCHEMA_VERSION {
+                log::warn!(
```
rfk, line 83: nit: seeing experiments with a higher major version number will at some point be an expected part of operation of this component, so I feel like this could be an `.info!` rather than a `.warn!`.

nimbus/src/client/http_client.rs
```diff
@@ -91,42 +113,75 @@ mod tests {
...
-        let body = r#"
-        { "data": [
-            {
+        // There are two experiments defined here, one has a "newer" schema version
+        // in order to test filtering of unsupported schema versions.
+        let body = format!(
+            r#"
+        {{ "data": [
+            {{
+                "schemaVersion": "{}.0.0",
```
rfk, line 122: Since the thing being formatted is such a long string, you might find it useful to use [named format arguments](https://doc.rust-lang.org/std/fmt/#named-parameters) here. Having this one as `"schemaVersion": "{current_version}.0.0"` and the other one as `"schemaVersion": "{newer_version}.0.0"` might make the test a little easier to understand for future readers.

nimbus/src/client/http_client.rs
```diff
@@ -148,6 +203,7 @@ mod tests {
         assert_eq!(
             exp.clone(),
             Experiment {
+                schema_version: "1.0.0".to_string(),
```
rfk, line 206: nit: since you're hard-coding the expected schema version here in the assertion, I think it would also be fine to hard-code the schema versions in the test data rather than formatting based on `SCHEMA_VERSION`. Alternately, this should also check for `format!("{}.0.0", SCHEMA_VERSION)` rather than hard-coding, for consistency.

nimbus/src/client/http_client.rs
```diff
@@ -66,6 +67,27 @@ impl SettingsClient for Client {
...
+                Some(ver) => {
+                    serde_json::from_value::<String>(ver.to_owned()).unwrap_or("".to_string())
+                }
+                None => {
+                    log::trace!("Missing schemaVersion: {:#?}", exp);
+                    continue;
+                }
+            };
+            let schema_maj_version = exp_schema_version.split(".").next().unwrap_or("");
+            let schema_version: i32 = FromStr::from_str(schema_maj_version).unwrap();
```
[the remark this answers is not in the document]
rfk: Oh, also, using `u32` rather than `i32` here would help guard against us accidentally accepting `-1.0.0` as a version number :-)

rfk (approved):
LGTM! I left a couple of minor nits about general Rust stuff, but the only thing I care deeply about is not panicking if the schema major version doesn't parse as an integer, and instead cleanly ignoring the experiment.

It may be useful to add a few more tests to cover the handling of malformed schema versions, but we haven't gone into that much depth in other parts of the codebase here so that's not a blocker.

PR #51, opened 2020-10-29
Add global opt in/out

jhugman:
This PR adds a global opt in or opt out for all experiments for the user to express a preference about participating in any experiments.

If the user has opted out, `get_active_experiments()` returns an empty list of experiments.

The naming is a bit clunky, because:

1. This flag codes for "don't enroll  the user in new experiments" as well as "opt out of already enrolled experiments".
2. I'd prefer it to be a positive choice (`!is_enabled` is better than `is_disabled`).

The name `globalUserParticipation` was first proposed in the [Nimbus Android Component API doc][1]

This corresponds to @tdsmith  [Unpacking opting out][3] compatible with [Intent to treat analysis][4].

 * If the user is already enrolled in an experiment, then the experiment is given an enrollment status of `OptedOut`.
 * When new experiments are encountered, no enrollment status is given.

Linked to [Nimbus SDK-36][2].

[1]: https://docs.google.com/document/d/1kchihPal0__A4VOAiPJarNuZXns5KhOHHfIeIzT6zfU/edit#
[2]: https://jira.mozilla.com/projects/SDK/issues/SDK-36
[3]: https://docs.google.com/document/d/18Vdn4zZX2D1u9AqzHRY0lUSUbpBdouqWtVp3Bu8vYQA/edit#
[4]: https://en.wikipedia.org/wiki/Intention-to-treat_analysis

### DIFF
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -28,6 +28,8 @@ use uuid::Uuid;
 
 const DEFAULT_TOTAL_BUCKETS: u32 = 10000;
 const DB_KEY_NIMBUS_ID: &str = "nimbus-id";
+const DB_KEY_GLOBAL_USER_PARTICIPATION: &str = "user-opt-in";
+const DEFAULT_GLOBAL_USER_PARTICIPATION: bool = true;
 
 /// Nimbus is the main struct representing the experiments state
 /// It should hold all the information needed to communicate a specific user's
@@ -80,7 +82,23 @@ impl NimbusClient {
             .map(|e| e.branch_slug.clone()))
     }
 
+    pub fn get_global_user_participation(&self) -> Result<Option<bool>> {
+        self.db()?.get(StoreId::Meta, DB_KEY_GLOBAL_USER_PARTICIPATION)
+    }
+
+    pub fn set_global_user_participation(&self, flag: bool) -> Result<()> {
+        self.db()?.put(StoreId::Meta, DB_KEY_GLOBAL_USER_PARTICIPATION, &flag)
+    }
+
+    fn is_enabled(&self) -> Result<bool> {
+        let user_choice = self.get_global_user_participation()?;
+        Ok(user_choice.unwrap_or(DEFAULT_GLOBAL_USER_PARTICIPATION))
+    }
+
     pub fn get_active_experiments(&self) -> Result<Vec<EnrolledExperiment>> {
+        if !self.is_enabled()? {
+            return Ok(Default::default());
+        }
         self.maybe_initial_experiment_fetch()?;
         get_enrollments(self.db()?)
     }
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -17,7 +17,8 @@ pub use evaluator::evaluate_enrollment;
 use client::{create_client, SettingsClient};
 pub use config::RemoteSettingsConfig;
 use enrollment::{
-    get_enrollments, opt_in_with_branch, opt_out, reset_enrollment, update_enrollments,
+    get_enrollments, get_global_user_participation, opt_in_with_branch, opt_out, reset_enrollment,
+    set_global_user_participation, update_enrollments,
 };
 pub use matcher::AppContext;
 use once_cell::sync::OnceCell;
@@ -28,8 +29,6 @@ use uuid::Uuid;
 
 const DEFAULT_TOTAL_BUCKETS: u32 = 10000;
 const DB_KEY_NIMBUS_ID: &str = "nimbus-id";
-const DB_KEY_GLOBAL_USER_PARTICIPATION: &str = "user-opt-in";
-const DEFAULT_GLOBAL_USER_PARTICIPATION: bool = true;
 
 /// Nimbus is the main struct representing the experiments state
 /// It should hold all the information needed to communicate a specific user's
@@ -82,23 +81,22 @@ impl NimbusClient {
             .map(|e| e.branch_slug.clone()))
     }
 
-    pub fn get_global_user_participation(&self) -> Result<Option<bool>> {
-        self.db()?.get(StoreId::Meta, DB_KEY_GLOBAL_USER_PARTICIPATION)
+    pub fn get_global_user_participation(&self) -> Result<bool> {
+        get_global_user_participation(self.db()?)
     }
 
     pub fn set_global_user_participation(&self, flag: bool) -> Result<()> {
-        self.db()?.put(StoreId::Meta, DB_KEY_GLOBAL_USER_PARTICIPATION, &flag)
-    }
-
-    fn is_enabled(&self) -> Result<bool> {
-        let user_choice = self.get_global_user_participation()?;
-        Ok(user_choice.unwrap_or(DEFAULT_GLOBAL_USER_PARTICIPATION))
+        set_global_user_participation(self.db()?, flag)?;
+        // Now update all enrollments based on this new opt in/out setting.
+        update_enrollments(
+            self.db()?,
+            &self.nimbus_id()?,
+            &self.available_randomization_units,
+            &self.app_context,
+        )
     }
 
     pub fn get_active_experiments(&self) -> Result<Vec<EnrolledExperiment>> {
-        if !self.is_enabled()? {
-            return Ok(Default::default());
-        }
         self.maybe_initial_experiment_fetch()?;
         get_enrollments(self.db()?)
     }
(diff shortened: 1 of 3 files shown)


nimbus/src/lib.rs
```diff
@@ -80,7 +82,23 @@ impl NimbusClient {
...
+    }
+
+    pub fn set_global_user_participation(&self, flag: bool) -> Result<()> {
+        self.db()?.put(StoreId::Meta, DB_KEY_GLOBAL_USER_PARTICIPATION, &flag)
+    }
+
+    fn is_enabled(&self) -> Result<bool> {
+        let user_choice = self.get_global_user_participation()?;
+        Ok(user_choice.unwrap_or(DEFAULT_GLOBAL_USER_PARTICIPATION))
+    }
```
jhugman, line 96: Unpacking whether the user has opted in or out, and how that actually affects the library is done in a private method so we can use it.

I'm not sure of the caching properties of `rkv`, so unsure whether it's worth storing a property on this `NimbusClient` struct itself.

nimbus/src/lib.rs
```diff
@@ -80,7 +82,23 @@ impl NimbusClient {
...
+
     pub fn get_active_experiments(&self) -> Result<Vec<EnrolledExperiment>> {
+        if !self.is_enabled()? {
+            return Ok(Default::default());
+        }
```
jhugman, line 101: Doing the check here means we won't report membership in `get_experiment_branch`.

We might also want to prevent the `update_experiments` to happen too.

jhugman (commented):
This needs tests, and wiring up the opt out to the command line app.

nimbus/src/lib.rs
```diff
@@ -80,7 +82,23 @@ impl NimbusClient {
             .map(|e| e.branch_slug.clone()))
     }
 
+    pub fn get_global_user_participation(&self) -> Result<Option<bool>> {
+        self.db()?.get(StoreId::Meta, DB_KEY_GLOBAL_USER_PARTICIPATION)
+    }
+
+    pub fn set_global_user_participation(&self, flag: bool) -> Result<()> {
+        self.db()?.put(StoreId::Meta, DB_KEY_GLOBAL_USER_PARTICIPATION, &flag)
```
mhammond, line 90: I imagine that setting this to false should "do something" too? I raised this in slack but it got lost in other noise, but we should possibly clarify the requirements here. The general question is "does opting out then opting back in do a full reset?" Specifically:

* On a global opt-out then opt back in, do I reuse my nimbus-id? If so, that implies I'll probably be re-enrolled in the exact same experiments.
* If I opt out of an experiment, then opt out globally, then opt in globally, am I again eligible for that first experiment?

nimbus/src/nimbus.idl
```diff
@@ -63,6 +63,17 @@ interface NimbusClient {
     [Throws=Error]
     sequence<EnrolledExperiment> get_active_experiments();
 
+    // Getter and setter for user's participation in all experiments.
+    // Possible values are:
+    // * `true`: user has opted in.
+    // * `false`: user has opted out
+    // * null: user has not made a choice.
```
rfk, line 70: Do we expect we'll want to reflect this `null` state in the UX differently at all?

mhammond, line 70: I don't feel strongly about this, but I don't really see how the `null` result is helpful - the SDK is assuming either true or false, and there's no way for callers to know. I don't see how UI could effectively represent this tri-state - everything needs to see this as a bool, so that 3rd state just seems to offer potential confusion.

(I guess it might make sense if there was some way for the app to dictate what the default is (eg, true on nightly, false on release or similar) it might make some sense.

mhammond (commented):
A quick drive-by

nimbus/src/lib.rs
travis79: For some context, Glean rotates it's client id every time the user opts out, so I am leaning towards rotating the `nimbus_id` for the reasons @mhammond mentions. Unfortunately, that might make it difficult to track that a user was enrolled but "opted out" of an experiment via telemetry, which sounds like it will be a requirement for analysis purposes.

jhugman: These are good points, both.

I'm not clear on the loose ends need tying up when we're telling anyone we're no longer enrolled in.

My understanding is that the `nimbus_id` is only used internally to nimbus, and never leaves the device.

 * what happens when the user opts out?
     - do we need to tell anyone/Glean that this device is no longer enrolled in any experiments? AFAICT this will be picked up next ping that is sent, that this `client_id` is enrolled in no experiments. Is this soon enough, or should we do that immediately?
 * what happens to when the user re-opts in? 
    - Do we need to tell anyone/Glean that this devices is now enrolled in a set of experiments? This is complicated by the fact that Glean and Nimbus don't meet anywhere where Nimbus can call Glean directly.
     - I suspect that the re-bucketing would not be desirable. From a UX point-of-view, the user would be rolling the dice of which feature they get each time they ticked and unticked a box. I don't know if there are subtle statistics problems if you don't do that rebucketing, though.

@mhammond @travis79 what say you?

travis79: I believe that the idea is to record these changes of state as events in Glean, so if the user opts out, or opts back in, we can record the exact moment and state.  In the case here, I don't think opting out and back in should land you in a different bucket or branch, in fact, I think that once a user has opted out, that they should not be allowed to opt back into the same experiment.  With the effect that this might have on analysis, it might be a good question to get data-science input on for what the desired behavior is.

nimbus/src/nimbus.idl
jhugman: This would be useful in asking if the user can opt-in to experiments. We don't want to keep asking the user to make a choice once they've made a positive or negative choice.

However, Kate's comment: 

> We have no plans to support real experiments that would require opt-in consent any time soon.

would suggest that we're not going to need such UI.

nimbus/src/lib.rs
mhammond: >  I don't think opting out and back in should land you in a different bucket or branch

That's a good point and would argue for keeping the same ID.

>  think that once a user has opted out, that they should not be allowed to opt back into the same experiment.

User's will not be able to opt-in - but as above, if we change their ID they might now qualify - IOW, changing their ID gives them a back-door into opting-in - they just keep opting out them back in again until their new ID qualifies them.

>  it might be a good question to get data-science input

Yes, exactly - I don't think our opinions are particularly relevant :)

rfk: > what happens when the user opts out? 

Tim's document on [unpacking unenrollment](https://docs.google.com/document/d/18Vdn4zZX2D1u9AqzHRY0lUSUbpBdouqWtVp3Bu8vYQA/) seems relevant here, particularly:

"""
Any client that enrolls in the experiment and is assigned to a branch should have their data included for the assigned condition in the final analysis, even if they withdraw and subsequently have a different user experience than they were assigned
"""

By this logic, if a user opts out *after having been enrolled in an experiment*, then we should continue to report that experiment and branch in their glean metrics until such time as the experiment ends. We should not actually show them the experimental treatment, because they've opted out of the experiment. But the fact that they were in the experiment at some point is relevant to analysis.  (And if they don't want us to report on this at all, they can disable telemetry).

Implicit in this requirement is that we remember a user's past enrollments, rather than deriving them afresh each time using the nimbus_id. So I don't think changing nimbus_id should have any effect on for experiment in which they'd previously been enrolled (and also, it's not clear to me what the intended effect of changing it would be).

nimbus/src/nimbus.idl
jhugman: For opt-in experiments this would be important, however, we're not intending to support opt-in experiments.

jhugman: Switched to a `true`/`false`.

nimbus/src/lib.rs
jhugman: > Implicit in this requirement is that we remember a user's past enrollments, rather than deriving them afresh each time using the nimbus_id. So I don't think changing nimbus_id should have any effect on for experiment in which they'd previously been enrolled (and also, it's not clear to me what the intended effect of changing it would be).

I think the `nimbus_id` is an implementation detail of how the enrollment occurs; as long as it's deterministic, I don't think we need to store the past enrollment.

If the experiment changes (e.g. branches are removed) while the user is opted out, and then the user opts back in, then we'll be in trouble either way: by remembering the enrollment the now defunct branch and reason, OR enrolling into a new branch, but I think this is for a different issue to solve.

rfk: > but I think this is for a different issue to solve.

Agreed

travis79 (approved):
LGTM!

PR #52, opened 2020-11-03
Generate schemas from NimbusExperiment structs

k88hudson:
I got this partially working, looking for feedback on: 
* does this crate structure make sense? i.e. the `generate` nested crate that we could theoretically call from CI?
* I couldn't figure out how to get `additionalProperties: true` working – any ideas there?
* Could we just use `NimbusExperiment` as-is in `nimbus`?


rfk:
Broadly, is the intention that this repo would become the canonical home for the schema definitions, derived from the Rust struct definitions?

> does this crate structure make sense? i.e. the generate nested crate that we could theoretically call from CI?

You may find it simpler to make this executable part of the `nimbus-shared` crate rather than as a separate sub-crate, especially since it doesn't appear to pull in any extra dependencies. There are some notes on that kind of setup [here](https://stackoverflow.com/questions/26946646/rust-package-with-both-a-library-and-a-binary), but broadly:

* Have `./src/lib.rs` for the library and `./src/main.rs` for the binary
* Add a `[[bin]]` section describing the executable to `nimbus-shared/Cargo.toml`.

It's OK for this executable to be published to crates.io, IIRC cargo is smart enough to build only the library target and not any executables when pulling in things from crates.io as dependencies.

> Could we just use NimbusExperiment as-is in nimbus?

It seems worth trying at least 👍

piatra:
At first I thought you *have* to type all your fields accordingly and can't do random additional properties but no, they just didn't document this in the package (at least not properly I had to read through their code).

One of the test expected outputs had `additionalProperties: true` and I traced that back to [the rust type](https://github.com/GREsau/schemars/blob/460976e765d08a02a7a3cd67b15c85107bb73fb0/schemars/src/schema.rs#L164-L168). The comment there kinda explains what's happening. It does add a dependency on `serde_json`.

This got it working for me:

```patch
diff --git a/nimbus-shared/src/lib.rs b/nimbus-shared/src/lib.rs
index 75540c6..6f853a9 100644
--- a/nimbus-shared/src/lib.rs
+++ b/nimbus-shared/src/lib.rs
@@ -1,4 +1,5 @@
-use schemars::JsonSchema;
+use schemars::{JsonSchema, Map};
+use serde_json::Value;

 pub const VERSION: &'static str = env!("CARGO_PKG_VERSION");

@@ -68,6 +69,9 @@ pub struct NimbusExperiment {
     /// Unique identifier for the experiment. This is a duplicate of slug, but is a required field
     /// for all Remote Settings records.
     pub id: String,
+
+    #[serde(flatten)]
+    pub extensions: Map<String, Value>,
 }
```

nimbus-shared/generate/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "generate"
+version = "0.1.0"
+authors = []
+# This is just for circle, so does not need to be published
+publish = false
+
+[dependencies]
+schemars = "0.8.0"
+serde_json = "1.0"
```
travis79, line 10: This dependency version of `"1.0"` doesn't exactly match up with the base Cargo.toml dependency on serde listed as `"1"`.  Not sure that would cause any major problems or not, but you may not quite resolve the version you expect in one or the other place.  If you make the changes suggested by @rfk for the structure, then this shouldn't be a problem anymore.

nimbus-shared/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+#[derive(JsonSchema)]
+#[allow(non_camel_case_types)]
+pub enum RandomizationUnit {
+    nimbus_id,
+    normandy_id,
+}
+
+#[derive(JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct NimbusExperiment {
```
travis79, line 14: Will this replace the definition of the `Experiment` in `./nimbus/lib.rs`?  If so, my PR (#49) may complicate things.  In order to support a range of versions for the backward-compatibility requirement, I changed `Experiment` to be an enum with variants representing the different schema versions.

It might not be so bad, as I should be able to add a variant that points to this representation, and then we can save the supported "historical" variants as structs named by their version or something.

PR #53, opened 2020-11-06
Release v0.3.0

rfk:
The rust crate claims to be at v1.0.0 already so I figured I'd make the Kotlin package match, but I'm open to bikeshedding on that front.

### DIFF
--- a/android/.buildconfig.yaml
+++ b/android/.buildconfig.yaml
@@ -1,4 +1,4 @@
-libraryVersion: 0.2.0
+libraryVersion: 0.3.0
 groupId: org.mozilla.experiments
 artifactId: nimbus
 description: A uniffi generated android bindings for the Nimbus SDK
--- a/nimbus/Cargo.toml
+++ b/nimbus/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "nimbus-sdk"
-version = "1.0.0"
+version = "0.3.0"
 authors = ["The Glean Team <glean-team@mozilla.com>", "The Sync Team <sync-team@mozilla.com>"]
 edition = "2018"
 description = "A rapid experiment library"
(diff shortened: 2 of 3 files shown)


rfk:
Hrm, actually I think we wanted to do something about the `viaduct` dependency here too, let me see...

nimbus/Cargo.toml
```diff
@@ -20,7 +20,7 @@ serde_derive = "1"
 serde_json = "1"
 log = "0.4"
 # Viaduct locked to version 61.0.10 of Application Services
-viaduct = { git = "https://github.com/mozilla/application-services",  rev = "2478bcf2b48d1867b01e8b7df4f86a69d564d49a"}
+viaduct = { git = "https://github.com/mozilla/application-services",  rev = "5bbda246ec2ef14c3e4ee6dd37541b038d63d6f6"}
```
rfk, line 23: This is the current head of appservices

dmose: I thought we wanted to largely unpin app-services?  Eg put it at 61 or whatever the current head is?

rfk: It's not released on crates.io, so we have to pin to *some* rev here, but we could use a tag rather than commit hash maybe?

dmose:
Given that Nimbus currently depends on a version of UniFFI that (among other things) generates GeckoJS bindings that pretty much always crash, it seems like 1.0.0 might send the wrong message about the maturity of the SDK...

That said, given the small number of consumers here, maybe it Just Doesn't Matter.

nimbus/Cargo.toml
dmose: That at least communicates more clearly what the point of the pin is (i.e. does it really require that _exact_ rev, or...).

dmose:
Re version 1.0, we have yet to run a real experiment (even A/A) on any platform...

android/.buildconfig.yaml
```diff
@@ -1,8 +1,8 @@
-libraryVersion: 0.2.0
+libraryVersion: 0.3.0
```
rfk, line 1: I went with `0.3.0` to a have something unambiguously different for the Kotlin and Rust.

nimbus/Cargo.toml
rfk: yep, I'm going to try with a tag, thanks

dmose (approved):
Looks good to me; r=@dmose

PR #54, opened 2020-11-06
Use a signed byte for the dummy value in AvailableRandomizationUnits.

rfk:
Unsigned types are currently experimental in Kotlin, and can only be
used by consumers if they explicitly opt in. Let's not put this opt-in
requirement on all consumers of this SDK if we can avoid it.

### DIFF
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -231,7 +231,7 @@ impl Default for RandomizationUnit {
 #[derive(Default)]
 pub struct AvailableRandomizationUnits {
     pub client_id: Option<String>,
-    dummy: u8, // See comments in nimbus.idl for why this hacky item exists.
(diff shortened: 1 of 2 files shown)


rfk:
@dmose requesting r? from you in case this change would affect the desktop work in unexpected ways.

dmose (approved):
I'm not aware of any issues that would cause (famous last words), especially since it's a dummy.  I'd be interested to hear if @eoger knows something that I don't, but I'll let you decide whether that should block this from landing.  r=dmose

mhammond:
FWIW, I chose `u8` at random here when adding it, so I don't see why `i8` isn't fine.

rfk:
> FWIW, I chose u8 at random here when adding it, so I don't see why i8 isn't fine.

In that case, let's ship it :-)

PR #58, opened 2020-11-06
Update to rkv v0.15

rfk:
Not sure if we have any particular concerns about this new version, but it's the version that glean is using so it seems worthwhile to try to stay aligned.

We could also consider copying some of the "safe-mode" infrastructure that Glean is using, which IIUC will be required for using this on Desktop. But that can be for a separate PR.

### DIFF
--- a/nimbus/Cargo.lock
+++ b/nimbus/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.13"
+version = "0.7.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


mhammond:
> We could also consider copying some of the "safe-mode" infrastructure that Glean is using, which IIUC will be required for using this on Desktop. But that can be for a separate PR.

Yes! I've a WIP for transactions that does what's here and what you describe, but let get this in now.

mhammond (approved):
(I'm not an sdk reviewer, but r+ anyway? :)

dmose:
Are there links to the safe-mode stuff and/or what Mark's doing?  @andrei and I will find it helpful to see what's coming down the pike for desktop...

PR #60, opened 2020-11-09
Update to uniffi v0.5.0

rfk:
This makes the crate compatible with Rust v1.43, which is required
for vendoring into application-services.


rfk:
(This PR obviously won't pass CI until `v0.5.0 is released, ref https://github.com/mozilla/uniffi-rs/pull/346)

travis79 (approved):

dmose:
Cargo.lock is still pointing to all the 0.4 versions of the uniffi* stuff.  I suspect now that uniffi 0.5 is on crates.io, running it again will update things.

dmose (approved):
Running Cargo.lock fixed things right up!  r=dmose, and I'm going to push it now so that I can update my patch stack.

PR #61, opened 2020-11-09
Implement transactions and the choice of the rkv back-end via features.

mhammond:
There's a bit of duplication here that isn't ideal, but I think this is fairly reasonable - happy to play with other approaches though.

### DIFF
--- a/nimbus/src/persistence.rs
+++ b/nimbus/src/persistence.rs
@@ -15,14 +15,123 @@ use rkv::StoreOptions;
 use std::fs;
 use std::path::Path;
 
+// Inspired by Glean - use a feature to choose between the backends.
+// Select the LMDB-powered storage backend when the feature is not activated.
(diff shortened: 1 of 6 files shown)


eoger (approved):
LGTM, I'm wondering is we are ever going to use the lmdb mode for rkv?

dmose (approved):
I don't think I really have enough context to give useful feedback here.  I eyeballed the patch, and nothing jumped out at me, so that's something... :-)

nimbus/src/persistence.rs
```diff
@@ -15,14 +15,123 @@ use rkv::StoreOptions;
 use std::fs;
 use std::path::Path;
 
+// Stolen from glean - use a feature to choose between the backends.
```
travis79, line 18: Glean freely gives ;)
```suggestion
// Inspired by glean - use a feature to choose between the backends.
```

dmose (approved):

PR #62, opened 2020-11-09
Pin viaduct to same version as in m-c

dmose:


travis79 (approved):

mhammond (approved):

rfk (approved):
WFM. When we repackage this as part of the appservices megazord we actually override this to force it to use the current head of appservices anyway, so I assume this will only affect things when vendored into m-c.

PR #65, opened 2020-11-16
Add a proof-of-concept Glean metric in the Rust code.

rfk:
What is says in the title. I'm opening this for early visibility and so I can easily find it again later, I don't expect we'll actually merge this.


nimbus/src/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+        // the application's Kotlin code, which uses its own separate dispatcher
+        // mechanism. Nothing tells the Rust code that Glean has been initialized
+        // externally, and the RLB dispatcher just keeps queueing up tasks and
+        // never actually dispatching them.
+        //
+        // We'll need to figure out an approach here with the Glean team.
+        // For now I'm lighty hacked up the Glean RLB code to expose this
+        // function that we can call to just pretend initialization was done
+        // and hope for the best...
+        let _ = glean::dispatcher::flush_init();
```
rfk, line 33: /cc @travis79 @badboy could I please get a gutcheck on my reasoning above?

From poking around in the Glean RLB code, it seems to use a background dispatcher thread for handling the actual setting of metrics. This dispatcher will queue up tasks until after the application calls `glean::initialize`.

For our use-case as a Rust component that gets used in some other application, the initialization happens in e.g. the Kotlin bindings rather than in RLB. So nothing ever calls `glean::initialize`, the dispatcher never executes any tasks, and the metrics emitted from Rust code never get sent off to Glean.

I've hacked around it here for a proof-of-concept and confirmed that metrics emitted from Rust do get sent once `flush_init` has been called, but this clearly isn't something we should ship. Any advice on what the right path forward here is?

Dexterp37: > 
> 
> /cc @travis79 @badboy could I please get a gutcheck on my reasoning above?

Hey @rfk , I'm joining the party here, too.

I believe you uncovered a big flaw that will need to be fixed before any integration in nimbus/sync could be done.

Please note that the RLB are **not yet meant to be used**, as they are actively being developed. The set of features we're developing are mainly to cover FOG immediate needs.

> For our use-case as a Rust component that gets used in some other application, the initialization happens in e.g. the Kotlin bindings rather than in RLB. So nothing ever calls `glean::initialize`, the dispatcher never executes any tasks, and the metrics emitted from Rust code never get sent off to Glean.

Yes, you are correct, that is a problem. The initialization part is easy to solve (e.g. we might have other bindings signal that init is complete), the bigger concern that I have lies in the order of the operations across the bindings.

Up to now, we only had one dispatcher to care about, and we relied on the fact that the dispatcher essentially acted as an off the main thread FIFO queue. We'll need to think carefully what having two dispatchers mean for the data guarantees that Glean has to offer or if there's any way to completely remove the Kotlin dispatcher and only use the Rust one.

> I've hacked around it here for a proof-of-concept and confirmed that metrics emitted from Rust do get sent once `flush_init` has been called, but this clearly isn't something we should ship. Any advice on what the right path forward here is?

I'd personally recommend not doing any integration right now and wait for the RLB to reach a good maturity stage :)

badboy: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1677438

dmose:
After reading through this, as well as Jan-Erik’s proposal in the Glean bug, I think Nimbus may also have an architectural issue somewhat analogous to this around being able to use both the C++ and JS bindings in Desktop for different experiments. They will presumably want to share a single NimbusClient instance, and given different thinking around the C++ and JS binding threading semantics, this could get interesting...

cc @piatra 

I’ll spin off a separate issue for this after the holidays...

nimbus/src/metrics.rs
mhammond: I'm still getting my head around the bigger picture here, so a naive question:

> Up to now, we only had one dispatcher to care about,

Is it necessary for this to mean 2 dispatchers, or can the RLB and Kotlin bindings arrange to share a dispatcher behind <handwave> some suitable abstraction?

FOG must have a roughly analogous issue with JS and the RLB in the same process - how is this handled there?

mhammond: > or can the RLB and Kotlin bindings arrange to

(lol - I had '\<handwave\>' there, but github stripped it :)

badboy: The [linked proposal](https://docs.google.com/document/d/1fc-WN5mEwgrJG9EB1ZupYTIAZchWTrm3JIGMVLcK6p4/edit) expands on this:

* Eventually we want to move the dispatcher back into Rust, so it's the single one used across all bindings, but that's a bit more work before we can get to it
* We can relax our guarantees, and that's already what we do on FOG, though there it's because of IPC.

IMO relaxing our guarantees is not too bad, so we're going ahead with that, but we still need to properly implement that (due to EoY and time-off I haven't moved forward on the proposal, but I think it's bascially accepted, so I'm going to file implemenation bugs and we'll get to it.)

PR #67, opened 2020-11-18
Don't implicitly update experiments on first use.

rfk:
(Based on the behaviour @eoger and @travis79 noted in our meeting earlier today).

Previously, we would implicity call `update_experiments` when trying
to list the experiments in an empty database. We would not update
if the database contained at least one experiment, no matter how stale.

In practice this seems to be a bit of a footgun, because it looks
to consumers like we'll automatically update definitions when
initialising the client, but this only works once.

Instead, let's require consumers to make an explicit call to
`update_experiments` whenever they want to update the experiments.

This will require us to do https://jira.mozilla.com/browse/SDK-121 on
the android-components side.

### DIFF
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -60,19 +60,6 @@ impl NimbusClient {
         })
     }
 
-    // This is a little suspect but it's not clear what the right thing is.
-    // Maybe it's OK we initially start with no experiments and just need to
-    // wait for the app to schedule a regular `update_experiments()` call?
-    // But for now, if we have no experiments we assume we have never
-    // successfully hit our server, so should do that now.
-    fn maybe_initial_experiment_fetch(&self) -> Result<()> {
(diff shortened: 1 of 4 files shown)


nimbus/src/lib.rs
```diff
@@ -60,19 +60,6 @@ impl NimbusClient {
         })
     }
 
-    // This is a little suspect but it's not clear what the right thing is.
-    // Maybe it's OK we initially start with no experiments and just need to
-    // wait for the app to schedule a regular `update_experiments()` call?
```
rfk, line 65: I think that not only is it *OK* to rely on the app to call `update_experiments`, it's well aligned with the overall philosophy of who should be in charge of dong what in this component.

travis79 (approved):
LGTM!

eoger (approved):
LGTM

PR #68, opened 2020-11-20
SDK-96 : Filter on application id

travis79:


nimbus/examples/experiment.rs
```diff
@@ -325,6 +328,7 @@ fn main() -> Result<()> {
                     .get_all_experiments()
                     .expect("can't fetch experiments!?")
                 {
+                    println!("{}", exp.slug.clone());
```
mhammond, line 331: Is this leftover by accident? If not, then adding a bit more context to what's printed would be good (and I'm surprised the clone is necessary?)

nimbus/src/enrollment.rs
```diff
@@ -351,6 +351,19 @@ mod tests {
...
+            app_build: None,
+            architecture: None,
+            device_manufacturer: None,
+            device_model: None,
+            locale: None,
+            os: None,
+            os_version: None,
+            android_sdk_version: None,
+            debug_tag: None,
+        };
```
mhammond, line 366:
```suggestion
        let app_context = AppContext {
            app_id: "fenix".to_string(),
            .. Default::default(),
        };
```

nimbus/src/enrollment.rs
```diff
@@ -407,11 +420,24 @@ mod tests {
         let db = Database::new(&tmp_dir)?;
         let nimbus_id = Uuid::new_v4();
         let aru = Default::default();
+        let app_context = AppContext {
+            app_id: "fenix".to_string(),
```
mhammond, line 424: (ditto here and below)

nimbus/src/evaluator.rs
```diff
@@ -45,7 +45,18 @@ pub fn evaluate_enrollment(
     app_context: &AppContext,
     exp: &Experiment,
 ) -> Result<ExperimentEnrollment> {
-    // Get targeting out of the way first - "if let chains" are experimental,
+    // Verify the application-id matches the application being targeted
+    // by the experiment.  This could also be done via the targeting
+    // but this is a short-circuit of sorts since the application field
```
mhammond, line 50: I agree this is better than doing it via `targeting` as it would mean an odd refactor - so I suggest you just delete that second sentence from the comment.

nimbus/src/evaluator.rs
```diff
@@ -323,42 +336,66 @@ mod tests {
...
-            &id,
-            &available_randomization_units,
-            &Default::default(),
-            &experiment1,
-        )
-        .unwrap();
+        // Application context for matching exp3
+        let context = AppContext {
+            app_id: "org.example.app".to_string(),
+            app_version: None,
```
mhammond, line 362: `.. Default::default()`?

mhammond (approved):
👍

nimbus/src/enrollment.rs
travis79: Ahh, that is much more concise 👍

nimbus/examples/experiment.rs
travis79: Just some debug garbage, nothing to see here...   Just forgot to remove it.

PR #69, opened 2020-11-27
Refactor enrollments and return events

eoger:
Fixes [SDK-122](https://jira.mozilla.com/browse/SDK-122)

This is quite a big patch, it implements the different enrolment state and transitions seen in [this document](https://docs.google.com/document/d/18L3oNriwwLd5TZrM1SaHIfIJXmXqUkwNwQEPwVOTY6Q/edit).

### DIFF
--- a/nimbus/src/enrollment.rs
+++ b/nimbus/src/enrollment.rs
@@ -1,23 +1,44 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
-use crate::error::Result;
 use crate::evaluator::evaluate_enrollment;
 use crate::persistence::{Database, StoreId, Writer};
+use crate::{error::Result, Error};
 use crate::{AppContext, AvailableRandomizationUnits, EnrolledExperiment, Experiment};
 
 use ::uuid::Uuid;
 use serde_derive::*;
-use std::collections::{HashMap, HashSet};
+use std::{
+    collections::{HashMap, HashSet},
+    time::{Duration, SystemTime, UNIX_EPOCH},
+};
 
 const DB_KEY_GLOBAL_USER_PARTICIPATION: &str = "user-opt-in";
 const DEFAULT_GLOBAL_USER_PARTICIPATION: bool = true;
+const PREVIOUS_ENROLLMENTS_GC_TIME: Duration = Duration::from_secs(30 * 24 * 3600);
 
 // These are types we use internally for managing enrollments.
 #[derive(Deserialize, Serialize, Debug, Clone, Hash, Eq, PartialEq)]
 pub enum EnrolledReason {
     Qualified, // A normal enrollment as per the experiment's rules.
     OptIn,     // Explicit opt-in.
+}
+
+// These are types we use internally for managing non-enrollments.
+#[derive(Deserialize, Serialize, Debug, Clone, Hash, Eq, PartialEq)]
+pub enum NotEnrolledReason {
+    OptOut,      // The user opted-out of experiments before we ever got enrolled to this one.
+    NotSelected, // The evaluator bucketing did not choose us.
+    NotTargeted, // We are not being targeted for this experiment.
+    EnrollmentsPaused, // The experiment enrollment is paused.
+}
+
+// These are types we use internally for managing disqualifications.
+#[derive(Deserialize, Serialize, Debug, Clone, Hash, Eq, PartialEq)]
+pub enum DisqualifiedReason {
+    Error,       // There was an error.
+    OptOut,      // The user opted-out from this experiment or experiments in general.
+    NotTargeted, // The targeting has changed for an experiment.
 }
 
 // Every experiment has an ExperimentEnrollment, even when we aren't enrolled.
@@ -27,21 +48,365 @@ pub struct ExperimentEnrollment {
     pub status: EnrollmentStatus,
 }
 
+impl ExperimentEnrollment {
+    /// Evaluate an experiment enrollment for an experiment
+    /// we are seeing for the first time.
+    fn from_new_experiment(
+        is_user_participating: bool,
+        nimbus_id: &Uuid,
+        available_randomization_units: &AvailableRandomizationUnits,
+        app_context: &AppContext,
+        experiment: &Experiment,
+        out_enrollment_events: &mut Vec<EnrollmentChangeEvent>,
+    ) -> Result<Self> {
+        Ok(if !is_user_participating {
+            Self {
+                slug: experiment.slug.clone(),
+                status: EnrollmentStatus::NotEnrolled {
+                    reason: NotEnrolledReason::OptOut,
+                },
+            }
+        } else if experiment.is_enrollment_paused {
+            Self {
+                slug: experiment.slug.clone(),
+                status: EnrollmentStatus::NotEnrolled {
+                    reason: NotEnrolledReason::EnrollmentsPaused,
+                },
+            }
+        } else {
+            let enrollment = evaluate_enrollment(
+                nimbus_id,
+                available_randomization_units,
+                app_context,
+                experiment,
+            )?;
+            log::debug!(
+                "Experiment '{}' is new - enrollment status is {:?}",
+                &enrollment.slug,
+                &enrollment
+            );
+            if matches!(enrollment.status, EnrollmentStatus::Enrolled { .. }) {
+                out_enrollment_events.push(enrollment.get_change_event())
+            }
+            enrollment
+        })
+    }
+
+    /// Force enroll ourselves in an experiment.
+    fn from_explicit_opt_in(
+        experiment: &Experiment,
+        branch_slug: &str,
+        out_enrollment_events: &mut Vec<EnrollmentChangeEvent>,
+    ) -> Result<Self> {
+        if !experiment.has_branch(branch_slug) {
+            return Err(Error::NoSuchBranch(
+                branch_slug.to_owned(),
+                experiment.slug.clone(),
+            ));
+        }
+        let enrollment = Self {
+            slug: experiment.slug.clone(),
+            status: EnrollmentStatus::new_enrolled(EnrolledReason::OptIn, branch_slug),
+        };
+        out_enrollment_events.push(enrollment.get_change_event());
+        Ok(enrollment)
+    }
+
+    /// Update our enrollment to an experiment we have seen before.
+    fn on_experiment_updated(
+        &self,
+        is_user_participating: bool,
+        nimbus_id: &Uuid,
+        available_randomization_units: &AvailableRandomizationUnits,
+        app_context: &AppContext,
+        updated_experiment: &Experiment,
+        out_enrollment_events: &mut Vec<EnrollmentChangeEvent>,
+    ) -> Result<Self> {
+        Ok(match self.status {
+            EnrollmentStatus::NotEnrolled { .. } => {
+                if !is_user_participating || updated_experiment.is_enrollment_paused {
+                    self.clone()
+                } else {
+                    let updated_enrollment = evaluate_enrollment(
+                        nimbus_id,
+                        available_randomization_units,
+                        app_context,
+                        updated_experiment,
+                    )?;
+                    log::debug!(
+                        "Experiment '{}' with enrollment {:?} is now {:?}",
+                        &self.slug,
+                        &self,
+                        updated_enrollment
+                    );
+                    if matches!(updated_enrollment.status, EnrollmentStatus::Enrolled { .. }) {
+                        out_enrollment_events.push(updated_enrollment.get_change_event());
+                    }
+                    updated_enrollment
+                }
+            }
+            EnrollmentStatus::Enrolled {
+                ref branch,
+                enrollment_id,
+                ..
+            } => {
+                if !is_user_participating {
+                    log::debug!(
+                        "Existing experiment enrollment '{}' is now disqualified (global opt-out)",
+                        &self.slug
+                    );
+                    let updated_enrollment = Self {
+                        slug: self.slug.clone(),
+                        status: EnrollmentStatus::Disqualified {
+                            reason: DisqualifiedReason::OptOut,
+                            enrollment_id,
+                            branch: branch.clone(),
+                        },
+                    };
+                    out_enrollment_events.push(updated_enrollment.get_change_event());
+                    updated_enrollment
+                } else if !updated_experiment.has_branch(branch) {
+                    // The branch we were in disappeared!
+                    let updated_enrollment = Self {
+                        slug: self.slug.clone(),
+                        status: EnrollmentStatus::Disqualified {
+                            reason: DisqualifiedReason::Error,
+                            enrollment_id,
+                            branch: branch.clone(),
+                        },
+                    };
+                    out_enrollment_events.push(updated_enrollment.get_change_event());
+                    updated_enrollment
+                } else {
+                    let evaluated_enrollment = evaluate_enrollment(
+                        nimbus_id,
+                        available_randomization_units,
+                        app_context,
+                        updated_experiment,
+                    )?;
+                    match evaluated_enrollment.status {
+                        EnrollmentStatus::Error { .. } => {
+                            let updated_enrollment = Self {
+                                slug: self.slug.clone(),
+                                status: EnrollmentStatus::Disqualified {
+                                    reason: DisqualifiedReason::Error,
+                                    enrollment_id,
+                                    branch: branch.clone(),
+                                },
+                            };
+                            out_enrollment_events.push(updated_enrollment.get_change_event());
+                            updated_enrollment
+                        }
+                        EnrollmentStatus::NotEnrolled {
+                            reason: NotEnrolledReason::NotTargeted,
+                        } => {
+                            log::debug!("Existing experiment enrollment '{}' is now disqualified (targeting change)", &self.slug);
+                            let updated_enrollment = Self {
+                                slug: self.slug.clone(),
+                                status: EnrollmentStatus::Disqualified {
+                                    reason: DisqualifiedReason::NotTargeted,
+                                    enrollment_id,
+                                    branch: branch.clone(),
+                                },
+                            };
+                            out_enrollment_events.push(updated_enrollment.get_change_event());
+                            updated_enrollment
+                        }
+                        EnrollmentStatus::NotEnrolled { .. }
+                        | EnrollmentStatus::Enrolled { .. }
+                        | EnrollmentStatus::Disqualified { .. }
+                        | EnrollmentStatus::WasEnrolled { .. } => self.clone(),
+                    }
+                }
+            }
+            EnrollmentStatus::Disqualified {
+                ref branch,
+                enrollment_id,
+                ..
+            } => {
+                if !is_user_participating {
+                    log::debug!(
+                        "Disqualified experiment enrollment '{}' has been reset to not-enrolled (global opt-out)",
+                        &self.slug
+                    );
+                    Self {
+                        slug: self.slug.clone(),
+                        status: EnrollmentStatus::Disqualified {
+                            reason: DisqualifiedReason::OptOut,
+                            enrollment_id,
+                            branch: branch.clone(),
+                        },
+                    }
+                } else {
+                    self.clone()
+                }
+            }
+            EnrollmentStatus::WasEnrolled { .. } | EnrollmentStatus::Error { .. } => self.clone(), // Cannot recover from errors!
+        })
+    }
+
+    /// Transition our enrollment to WasEnrolled (Option::Some) or delete it (Option::None)
+    /// after an experiment has disappeared from the server.
+    ///
+    /// If we transitioned to WasEnrolled, our enrollment will be garbage collected
+    /// from the database after `PREVIOUS_ENROLLMENTS_GC_TIME`.
+    fn on_experiment_ended(
+        &self,
+        out_enrollment_events: &mut Vec<EnrollmentChangeEvent>,
+    ) -> Option<Self> {
+        log::debug!(
+            "Experiment '{}' vanished while we had enrollment status of {:?}",
+            self.slug,
+            self
+        );
+        let (branch, enrollment_id) = match self.status {
+            EnrollmentStatus::Enrolled {
+                ref branch,
+                enrollment_id,
+                ..
+            } => (branch, enrollment_id),
+            EnrollmentStatus::Disqualified {
+                ref branch,
+                enrollment_id,
+                ..
+            } => (branch, enrollment_id),
+            EnrollmentStatus::NotEnrolled { .. }
+            | EnrollmentStatus::WasEnrolled { .. }
+            | EnrollmentStatus::Error { .. } => return None, // We were never enrolled anyway, simply delete the enrollment record from the DB.
+        };
+        let enrollment = Self {
+            slug: self.slug.clone(),
+            status: EnrollmentStatus::WasEnrolled {
+                enrollment_id,
+                branch: branch.to_owned(),
+                experiment_ended_at: now_secs(),
+            },
+        };
+        out_enrollment_events.push(enrollment.get_change_event());
+        Some(enrollment)
+    }
+
+    /// Force unenroll ourselves from an experiment.
+    fn on_explicit_opt_out(
+        &self,
+        out_enrollment_events: &mut Vec<EnrollmentChangeEvent>,
+    ) -> Result<Self> {
+        Ok(match self.status {
+            EnrollmentStatus::Enrolled {
+                enrollment_id,
+                ref branch,
+                ..
+            } => {
+                let enrollment = Self {
+                    slug: self.slug.to_string(),
+                    status: EnrollmentStatus::Disqualified {
+                        reason: DisqualifiedReason::OptOut,
+                        branch: branch.to_owned(),
+                        enrollment_id,
+                    },
+                };
+                out_enrollment_events.push(enrollment.get_change_event());
+                enrollment
+            }
+            EnrollmentStatus::NotEnrolled { .. } => Self {
+                slug: self.slug.to_string(),
+                status: EnrollmentStatus::NotEnrolled {
+                    reason: NotEnrolledReason::OptOut, // Explicitly set the reason to OptOut.
+                },
+            },
+            EnrollmentStatus::Disqualified { .. }
+            | EnrollmentStatus::WasEnrolled { .. }
+            | EnrollmentStatus::Error { .. } => {
+                // Nothing to do here.
+                self.clone()
+            }
+        })
+    }
+
+    /// Garbage collect old experiments we've kept a WasEnrolled enrollment from.
+    /// Returns Option::None if the enrollment should be nuked from the db.
+    fn maybe_garbage_collect(&self) -> Option<Self> {
+        if let EnrollmentStatus::WasEnrolled {
+            experiment_ended_at,
+            ..
+        } = self.status
+        {
+            let time_since_transition = Duration::from_secs(now_secs() - experiment_ended_at);
+            if time_since_transition < PREVIOUS_ENROLLMENTS_GC_TIME {
+                return Some(self.clone());
+            }
+        }
+        log::debug!("Garbage collecting enrollment '{}'", self.slug);
+        None
+    }
+
+    // Create a telemetry event describing the transition
+    // to the current enrollment state.
+    fn get_change_event(&self) -> EnrollmentChangeEvent {
+        match &self.status {
+            EnrollmentStatus::Enrolled {
+                enrollment_id,
+                branch,
+                ..
+            } => EnrollmentChangeEvent::new(
+                &self.slug,
+                enrollment_id,
+                branch,
+                None,
+                EnrollmentChangeEventType::Enrollment,
+            ),
+            EnrollmentStatus::WasEnrolled {
+                enrollment_id,
+                branch,
+                ..
+            } => EnrollmentChangeEvent::new(
+                &self.slug,
+                &enrollment_id,
+                &branch,
+                None,
+                EnrollmentChangeEventType::Unenrollment,
+            ),
+            EnrollmentStatus::Disqualified {
+                enrollment_id,
+                branch,
+                reason,
+                ..
+            } => EnrollmentChangeEvent::new(
+                &self.slug,
+                &enrollment_id,
+                &branch,
+                match reason {
+                    DisqualifiedReason::NotTargeted => Some("targeting"),
+                    DisqualifiedReason::OptOut => Some("optout"),
+                    DisqualifiedReason::Error => Some("error"),
+                },
+                EnrollmentChangeEventType::Disqualification,
+            ),
+            EnrollmentStatus::NotEnrolled { .. } | EnrollmentStatus::Error { .. } => unreachable!(),
+        }
+    }
+}
+
 #[derive(Deserialize, Serialize, Debug, Clone, Hash, Eq, PartialEq)]
 pub enum EnrollmentStatus {
-    // Enrolled
     Enrolled {
+        enrollment_id: Uuid, // Random ID used for telemetry events correlation.
         reason: EnrolledReason,
         branch: String,
     },
-    // Not enrolled because our evaluator declined to choose us.
-    NotSelected,
-    // Not enrolled because either enrollment is paused or the experiment is over.
-    NotRunning,
-    // Not enrolled because we are not being targeted for this experiment.
-    NotTargeted,
-    // User explicitly opted out.
-    OptedOut,
+    NotEnrolled {
+        reason: NotEnrolledReason,
+    },
+    Disqualified {
+        enrollment_id: Uuid,
+        reason: DisqualifiedReason,
+        branch: String,
+    },
+    WasEnrolled {
+        enrollment_id: Uuid,
+        branch: String,
+        experiment_ended_at: u64, // unix timestamp in sec, used to GC old enrollments
+    },
     // There was some error opting in.
     Error {
         // Ideally this would be an Error, but then we'd need to make Error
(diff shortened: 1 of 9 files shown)


nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
+    }
+}
+
+pub enum EnrollmentChangeEventType {
+    Enrollment,
+    Disqualification,
+    Unenrollment,
+}
+
+impl ExperimentEnrollment {
```
eoger, line 81: The meat of the patch is in this `impl`.

nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
+    },
+    Disqualified {
+        reason: DisqualifiedReason,
+        enrollment_id: String,
+        branch: String,
+    },
+    WasEnrolled {
+        enrollment_id: String,
+        branch: String,
+        observed_at: u64, // unix timestamp in sec, used to GC old enrollments
```
eoger, line 318: Note: Not implemented yet, that would be a follow-up to this patch.

nimbus/src/enrollment.rs
```diff
@@ -85,90 +365,156 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
...
+        Ok((updated_enrollments, enrollment_events))
+    }
+
+    fn evolve_enrollment(
+        &self,
+        is_user_participating: bool,
+        existing_experiment: Option<&Experiment>,
+        updated_experiment: Option<&Experiment>,
+        existing_enrollment: Option<&ExperimentEnrollment>,
+        enrollment_events: &mut Vec<EnrollmentChangeEvent>,
```
eoger, line 463: We use a `out` parameter to return the transition events.

nimbus/src/evaluator.rs
```diff
@@ -55,47 +57,43 @@ pub fn evaluate_enrollment(
             });
         }
     }
-    let status = if exp.is_enrollment_paused {
```
eoger, line 58: `is_enrollment_paused` is handled in the caller.

rfk:
(tagging myself for review, I'll try to take a look at this tomorrow)

nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
 pub struct ExperimentEnrollment {
     pub slug: String,
     pub status: EnrollmentStatus,
 }
 
+pub struct EnrollmentChangeEvent {
+    pub experiment_slug: String,
+    pub branch_slug: String,
+    pub enrollment_id: String,
+    pub reason: Option<String>,
```
mhammond, line 53: `reason` as a string seems odd - there are 2 arbitrary strings used in cases that don't make total sense to me - ie, why not a non-optional enum?

nimbus/src/enrollment.rs
mhammond, line 81: more a style question than a nit: it seems a bit odd to me that when the struct is in the same file that the impl doesn't immediately follow. I guess it might make sense if the "rule" as "all decls first, followed by impls", but that doesn't seem to be the case here either (ie, there's an `enum` following this impl)

nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
+                        .as_secs()
+                },
+            },
+        })
+    }
+}
+
 #[derive(Deserialize, Serialize, Debug, Clone, Hash, Eq, PartialEq)]
 pub enum EnrollmentStatus {
     // Enrolled
```
mhammond, line 301: nit: remove this comment too?

nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
     Enrolled {
+        enrollment_id: String, // Random ID used for telemetry events correlation.
```
mhammond, line 303: why not a `Uuid` (here and below)? Is it due to uniffi? If so, we have that problem in forms in a-s - so maybe a comment here, so it's obvious it should change should uniffi get better support?

nimbus/src/enrollment.rs
mhammond: I assume `observed_at` is when we observed we were no longer enrolled? Would changing the name to make that clearer be worthwhile? (OTOH, if it's not implemented, then there's probably a case to be made to just delete it?

nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
+                    Self {
+                        slug: self.slug.clone(),
+                        status: EnrollmentStatus::Disqualified {
+                            reason: DisqualifiedReason::OptOut,
+                            enrollment_id,
+                            branch,
+                        },
+                    }
+                } else {
+                    let mut updated_enrollment = evaluate_enrollment(
```
mhammond, line 197: Reading Ryan's document, we need to handle at least the fact that the experiment may now be closed for new enrollments, but we still need to stay enrolled. It's not clear to me from that doc what should happen if targeting changed, but I *think* the intent is that the original enrollment remains in that case too?

nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
+                }
+            }
+            EnrollmentStatus::Disqualified { .. } => {
+                if !is_user_participating {
+                    log::debug!(
+                        "Disqualified experiment enrollment '{}' has been discarded (global opt-out)",
+                        &self.slug
+                    );
+                    // Discard disqualified enrollments on global experiments opt-out.
+                    return Ok(None);
```
mhammond, line 245: Why is this case `None` rather than `EnrollmentStatus::NotEnrolled {reason: NotEnrolledReason::OptOut...}`, like is done in the fn above? Doesn't this mean the experiment will no longer exist in the DB?

nimbus/src/enrollment.rs
```diff
@@ -436,81 +842,104 @@ mod tests {
...
-        // // pretend we just updated from the server and one of the 2 is missing.
-        // db.delete(StoreId::Experiments, "secure-gold")?;
-        // update_enrollments(&db, &nimbus_id, &aru, &Default::default())?;
-        // // should only have 1 now.
-        // let enrollments = get_enrollments(&db)?;
-        // assert_eq!(enrollments.len(), 1);
+        assert_eq!(disqualified_enrollments.len(), 2);
         Ok(())
     }
 }
```
mhammond, line 945: I think some of the transitions aren't well covered here?

nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
+    }
+    fn on_experiment_updated(
+        &self,
+        is_user_participating: bool,
+        nimbus_id: &Uuid,
+        available_randomization_units: &AvailableRandomizationUnits,
+        app_context: &AppContext,
+        updated_experiment: &Experiment,
+        enrollment_events: &mut Vec<EnrollmentChangeEvent>,
+    ) -> Result<Option<Self>> {
```
mhammond, line 134: If I'm reading the update code correctly, returning `None` here will result in the experiment not existing in the db?

mhammond (commented):
This is a big patch indeed - looks fairly good, but I have a few nits and comments. I didn't quite finish the review but thought I'd get this up anyway.

nimbus/src/enrollment.rs
```diff
@@ -179,55 +525,117 @@ pub fn reset_enrollment(
...
 }
 
-pub fn opt_out(db: &Database, experiment_slug: &str) -> Result<()> {
-    // As above - check experiment exists?
+pub fn opt_out(db: &Database, experiment_slug: &str) -> Result<Vec<EnrollmentChangeEvent>> {
+    let mut writer = db.write()?;
+    let exp_store = db.get_store(StoreId::Experiments);
+    let _exp = match exp_store.get::<Experiment>(&writer, experiment_slug)? {
+        None => {
+            // XXX - do we want specific errors for this kind of thing?
```
piatra, line 618: Current JS implementation sends an `unenrollFailed` event
https://searchfox.org/mozilla-central/rev/66547980e8e8ca583473c74f207cae5bac1ed541/toolkit/components/messaging-system/experiments/ExperimentManager.jsm#285

nimbus/src/enrollment.rs
```diff
@@ -179,55 +525,117 @@ pub fn reset_enrollment(
...
+        Some(e) => e,
+    };
     let enrollment = ExperimentEnrollment {
         slug: experiment_slug.to_string(),
-        status: EnrollmentStatus::OptedOut,
+        status: EnrollmentStatus::Disqualified {
+            reason: DisqualifiedReason::OptOut,
+            // TODO: We should use the branch we're enrolled in if we had an enrollment
+            // But what do we do if we did not? I assume we want to support the "opt out
+            // before enrollment ever happens use case". Should we just make "branch" optional?
```
piatra, line 630: This currently does not happen in JS Desktop Nimbus. We check the `bucketConfig`/sampling to see if the user qualifies and then call `enroll` which [we are certain will succeed and choose a branch](https://searchfox.org/mozilla-central/rev/66547980e8e8ca583473c74f207cae5bac1ed541/toolkit/components/messaging-system/experiments/ExperimentManager.jsm#120-123).
And then on unenrollment we send the branch [in the event](https://searchfox.org/mozilla-central/rev/66547980e8e8ca583473c74f207cae5bac1ed541/toolkit/components/messaging-system/experiments/ExperimentManager.jsm#301).

nimbus/src/enrollment.rs
```diff
@@ -85,90 +365,156 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
...
-    app_context: &AppContext,
-) -> Result<()> {
-    log::info!("updating enrollments...");
-    // We might have enrollments for experiments which no longer exist, so we
-    // first build a set of all IDs in both groups.
-    let mut all_slugs = HashSet::new();
-    let experiments = db
-        .get_store(StoreId::Experiments)
-        .collect_all::<Experiment>(writer)?;
+pub(crate) struct EnrollmentsMerger<'a> {
```
rfk, line 368: FWIW, the significance of the name "enrollments merger" is not very clear to me - "merger" suggests that it's going to try to merge and reconcile some state, but what it actually does is evolve from one state to another. It's clunky but "Evolver" might be a better name than "Merger"?

nimbus/src/enrollment.rs
```diff
@@ -51,8 +326,13 @@ pub enum EnrollmentStatus {
 }
 
 impl EnrollmentStatus {
-    // This is used in examples, but not in the main dylib.
-    #[allow(dead_code)]
+    pub fn enrolled(reason: EnrolledReason, branch: &str) -> Self {
```
rfk, line 329: nit: could something like `new_enrolled` make it slightly more clear that this is a constructor?

nimbus/src/enrollment.rs
```diff
@@ -85,90 +365,156 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
...
+            app_context,
+        }
+    }
+
+    /// Convenient wrapper around `evolve_enrollments` that fetches the current state of experiments,
+    /// enrollments and user participation from the database.
+    pub(crate) fn evolve_enrollments_in_db(
+        &self,
+        db: &Database,
+        writer: &mut Writer,
```
rfk, line 392: To check my understanding - does the use of `writer` here mean this all runs in the context of a single transaction?

nimbus/src/enrollment.rs
rfk: As above, this should probably be a comment in the source itself. But it's also not entirely clear to me what we do this rather than returning a tuple from this function. (Especially since IIUC, each call to `evolve_enrollment` should result in at most one enrollment change event).

nimbus/src/enrollment.rs
```diff
@@ -179,55 +525,117 @@ pub fn reset_enrollment(
...
+    };
     let enrollment = ExperimentEnrollment {
         slug: experiment_slug.to_string(),
-        status: EnrollmentStatus::OptedOut,
+        status: EnrollmentStatus::Disqualified {
+            reason: DisqualifiedReason::OptOut,
+            // TODO: We should use the branch we're enrolled in if we had an enrollment
+            // But what do we do if we did not? I assume we want to support the "opt out
+            // before enrollment ever happens use case". Should we just make "branch" optional?
+            branch: "not-specified".to_owned(),
```
rfk, line 631: If we weren't enrolled, I think we should stay in the `NotEnrolled` state, and maybe set reason=`OptOut`. The `Disqualified` state should only be for tracking a particular branch assignment after being enrolled.

rfk (commented):
I haven't finished going through this, but need to put it down for some meetings, and wanted to push my first batch of comments in the meantime.

nimbus/src/enrollment.rs
```diff
@@ -85,90 +365,156 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
...
+                    existing_enrollment.on_experiment_updated(
+                        is_user_participating,
+                        self.nimbus_id,
+                        self.available_randomization_units,
+                        self.app_context,
+                        updated_experiment,
+                        enrollment_events,
+                    )?
+                }
+                (None, None, Some(_)) => unreachable!("Orphan enrollment."),
```
rfk, line 491: I'm not convinced that this state is unreachable, as we could have a `WasEnrolled` state that has been kept around after an experiment has ended. At the point in time when the experiment ends, we'd make a transition like:

```
(Some(_), None, Some(Enrolled(...))) => Some(WasEnrolled(...))
```

And then on some future update, we'll slurp this `WasEnrolled` state out of the db and process another transition like:

```
(None, None, Some(WasEnrolled(...))) => panic!
```

IIUC, this seems like the place where we want to garbage-collect old `WasEnrolled` states once sufficient time has passed. So instead of panicing, this should inspect the timestamp and either return `None` or `Some(WasEnrolled(...))`.

nimbus/src/enrollment.rs
```diff
@@ -85,90 +365,156 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
...
+                (None, Some(_), Some(_)) => {
+                    unreachable!("New experiment but enrollment already exists.")
+                }
+                (Some(_), None, None) | (Some(_), Some(_), None) => {
+                    unreachable!("Every experiment in the DB has an associated enrollment record.")
+                }
```
rfk, line 497: These are important invariants which feel like they should be documented somewhere, perhaps in the persistence layer schema documentation:

* Every known experiment should have a recorded enrollment.
* Enrollments records for unknown experiments should only ever be `WasEnrolled`.

I also wonder whether it's worth adding some asserts to enforce these, e.g. you could do a consistency check at the end of `evolve_enrollments_in_db` to make sure we're not putting the db into a bad state.

(An alternative way of enforcing these invariants, would be to store the experiment info as a field on the enrollment record, so that you can never store one without its corresponding other; not sure that would be worth the storage schema migration though)

nimbus/src/enrollment.rs
```diff
@@ -85,90 +365,156 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
...
+                        self.nimbus_id,
+                        self.available_randomization_units,
+                        self.app_context,
+                        updated_experiment,
+                        enrollment_events,
+                    )?
+                }
+                (None, None, Some(_)) => unreachable!("Orphan enrollment."),
+                (None, Some(_), Some(_)) => {
+                    unreachable!("New experiment but enrollment already exists.")
```
rfk, line 493: Since we have an `InternalError` type, I wonder if it would be better to return it here rather than panicking (and potentially crashing the entire app along with it).

nimbus/src/enrollment.rs
```diff
@@ -179,55 +525,117 @@ pub fn reset_enrollment(
...
+        .evolve_enrollment(
+            is_user_participating,
+            None, // Treat the experiment as we're seeing it for the first time...
+            Some(&exp),
+            None, // ...and we've never been enrolled to it.
+            &mut enrollment_events,
+        )?
+        .ok_or_else(|| {
+            Error::InternalError("reset_enrollment should have resulted in a non-empty enrollment")
+        })?;
```
rfk, line 551: I'm a little concerned about this, but I don't know enough context to comment for sure. Under what circumstances will the consuming app be expected to call this method?

The docstring suggests that it is for undoing an opt-out, but:

* I'm not aware of any product requirement to let to user opt back in to existing experiments after opting out.
  * When this came up in a recent Nimbus Architecture Group meeting, folks seemed pretty happy with the simplification of declaring this impossible, ref [Nov 12 meeting minutes](https://docs.google.com/document/d/12ARuGAa3vXrtS8f7_weg7oUAItREsRr7Vk0fV07JKhc/edit#heading=h.pz2i4nsa829b).
* It can reset much more than just opt-out status, e.g. if you were disqualified for some other reason then this could re-qualify you, or if you were in one branch and your context has changed then you could end up in a different branch.

This method feels like a footgun to me, and this patch is making the footguniness more obvious because of how it lets you make otherwise invalid state transitions. ISTM we should either limit it to just the specifically documented effects (e.g. only resetting states with `reason: OptOut`) or remove it entirely.

@mhammond maybe you can add more context here?

nimbus/src/enrollment.rs
```diff
@@ -179,55 +525,117 @@ pub fn reset_enrollment(
...
+        );
+        return Ok(vec![]);
+    }
     let enrollment = ExperimentEnrollment {
         slug: experiment_slug.to_string(),
-        status: EnrollmentStatus::Enrolled {
-            reason: EnrolledReason::OptIn,
-            branch: branch.to_string(),
-        },
+        status: EnrollmentStatus::enrolled(EnrolledReason::OptIn, branch),
```
rfk, line 593: Remind me, is this only for testing purposes, or do we expect consuming apps to expose some sort of opt-in UI for end users?

nimbus/src/enrollment.rs
```diff
@@ -179,55 +525,117 @@ pub fn reset_enrollment(
...
+    };
+    let enrollment_id = if let EnrollmentStatus::Enrolled { enrollment_id, .. } = &enrollment.status
+    {
+        enrollment_id
+    } else {
+        unreachable!(); // This is really dumb.
```
rfk, line 599: A `get_enrollment_id` helper method on `ExperimentEnrollment` might help clean this up a bit.

nimbus/src/enrollment.rs
rfk: To clarify: we should match on the current enrollment state here:

* If `Enrolled`, move to `Disqualified` with reason `OptOut` and the enrolled branch info.
* If `NotEnrolled`, stay in `NotEnrolled` but set the reason to `OptOut`.
* If `Disqualified` or `WasEnrolled`, don't change anything.

nimbus/src/enrollment.rs
```diff
@@ -436,81 +842,104 @@ mod tests {
...
-        // We should see no experiment enrolments.
-        assert_eq!(get_experiment_enrollments(&db)?.len(), 0);
-        assert_eq!(
-            get_experiment_enrollments_with_status(&db, EnrollmentStatus::OptedOut)?.len(),
-            0
-        );
+        assert_eq!(events.len(), 0);
+        // We should see the experiment non-enrollments.
+        assert_eq!(get_experiment_enrollments(&db)?.len(), 2);
+        let disqualified_enrollments: Vec<ExperimentEnrollment> = get_experiment_enrollments(&db)?
```
rfk, line 862: The name "disqualified enrollments" doesn't seem right to me here, you're filtering for `NotEnrolled` state rather than `Disqualified`.

nimbus/src/enrollment.rs
```diff
@@ -436,81 +842,104 @@ mod tests {
...
+        // We should see 2 experiment enrollments.
         assert_eq!(get_experiment_enrollments(&db)?.len(), 2);
-        assert_eq!(
-            get_experiment_enrollments_with_status(&db, EnrollmentStatus::OptedOut)?.len(),
-            0
-        );
-        let branches: Vec<String> = enrollments
-            .iter()
-            .map(|exp| exp.branch_slug.clone())
+        let disqualified_enrollments: Vec<ExperimentEnrollment> = get_experiment_enrollments(&db)?
```
rfk, line 888: As above, the name "disqualified enrollments" doesn't seem right here.

nimbus/src/enrollment.rs
```diff
@@ -85,90 +365,156 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
...
+    }
+
+    fn evolve_enrollment(
+        &self,
+        is_user_participating: bool,
+        existing_experiment: Option<&Experiment>,
+        updated_experiment: Option<&Experiment>,
+        existing_enrollment: Option<&ExperimentEnrollment>,
+        enrollment_events: &mut Vec<EnrollmentChangeEvent>,
+    ) -> Result<Option<ExperimentEnrollment>> {
```
rfk, line 464: I like the abstraction level of `evolve_enrollment` here, it feels like the sort of thing for which we could have a bit old suite of tests for all the various combos of things that change on the server and assert that the appropriate change happens on the client.

nimbus/src/enrollment.rs
rfk: Broadly, I don't have a good mental model for when this `opt_in` method will be called, so it's hard for me to think about edge-cases. What should happen if the user is already enrolled in a branch, should we violate our "never changes branches" rule here?

nimbus/src/enrollment.rs
```diff
@@ -179,55 +525,117 @@ pub fn reset_enrollment(
...
+            branch: "not-specified".to_owned(),
+            enrollment_id: "todo".to_owned(),
+        },
     };
-    let mut writer = db.write()?;
     db.get_store(StoreId::Enrollments)
         .put(&mut writer, experiment_slug, &enrollment)?;
     writer.commit()?;
-    Ok(())
+    Ok(vec![]) // TODO: we should probably emit a "disqualified" event depending on the previous enrollment status
```
rfk, line 638: I like the way that `evolve_enrollment` delegates to the `ExperimentEnrollment` via a kind of event-handler interface, and I wonder if leaning into that even more could help clean up these other state-changing functions. For example, rather than creating the new `ExperimentEnrollment` directly, you can imagine this calling an `existing_enrollment.on_opt_out` method that transitions to the appropriate new state, maintains any existing enrollment id, and returns a change event.

Likewise, you can imagine `on_opt_in` and `on_reset` methods that could be used by the functions above.

I think this would help us encapsulate the state-to-state transitions and make it obvious where and how to guard against illegal transitions.

nimbus/src/enrollment.rs
mhammond: I agree this should be removed as a public function. This is also called from `update_enrollments` to handle the case where the *global* opt-out state changes (and that does indeed only do it for the `OptOut` reason.. The meeting docs you linked above don't make it clear whether that decision also applies to the global state or not. If it does, then I agree it can be removed entirely.

rfk: > The meeting docs you linked above don't make it clear whether that decision also applies to the global state or not

Yes, my understanding is this should also apply to globally opting-out and then globally opting back in - you become eligible for new experiments, but if you were disqualified from an existing experiment then you stay disqualified.

nimbus/src/enrollment.rs
eoger: I implemented what was written in the document: `To Discarded state, when the user entirely disables telemetry in their client.`, but I think I agree with you: if you look closely at the `from_new_experiment` and `on_experiment_updated` (if `self.status == EnrollmentStatus::NotEnrolled`) function bodies, these 2 blocks do functionally the same thing.

nimbus/src/enrollment.rs
eoger: Yes I believe so!

nimbus/src/enrollment.rs
eoger: Good catch, and that answers https://github.com/mozilla/nimbus-sdk/pull/69#discussion_r532981366.

nimbus/src/enrollment.rs
rfk: Note that "when the user entirely disables telemetry" is different from "when the user opts out of experiments" and I'm not sure what API surface we have for the application to indicate this state.

nimbus/src/enrollment.rs
eoger: I agree it's awkward but originally I was returning tuples and the code looked even worse

rfk: Fair enough. I don't object to it, but we should try to make it more obvious in the code. One option if an inline comment seems unwieldy, would be to call the argument something like `enrollment_events_accumulator` or similar to flag how it's being used.

> I expect this to be temporary anyway

Haha, now that you've said it out loud, of course this code will be with us indefinitely :-)

eoger:
Pushed a new version with most of the comments addressed and a ton of tests for state transitions.

nimbus/src/enrollment.rs
```diff
@@ -20,28 +24,367 @@ pub enum EnrolledReason {
...
+            }
+            EnrollmentStatus::Disqualified { .. } => {
+                if !is_user_participating {
+                    log::debug!(
+                        "Disqualified experiment enrollment '{}' has been reset to not-enrolled (global opt-out)",
+                        &self.slug
+                    );
+                    Self {
+                        slug: self.slug.clone(),
+                        status: EnrollmentStatus::NotEnrolled {
```
rfk, line 254: They should stay in `Disqualified` but have `reason` set to `DisqualifiedReason::OptOut`, similar to what happens when we discover `!is_user_participating` while in the `Enrolled` state.  (Otherwise, the first transition after opting out would transition them from `Enrolled` to `Disqualified`, and the next one would take them from `Disqualified` to `NotEnrolled` and throw away their branch information).

nimbus/src/enrollment.rs
```diff
@@ -20,28 +24,367 @@ pub enum EnrolledReason {
...
+                        app_context,
+                        updated_experiment,
+                    )?;
+                    match updated_enrollment.status {
+                        EnrollmentStatus::Error { .. } => updated_enrollment,
+                        EnrollmentStatus::Enrolled {
+                            branch: ref new_branch,
+                            ..
+                        } if new_branch == branch => self.clone(),
+                        _ => {
```
rfk, line 222: IIUC this will disqualify the user with `DisqualifiedReason::NotTargeted` under any of the following circumstances:

* The targeting expression no longer matches the user
    * (here `status` would match `EnrollmentStatus::NotEnrolled { reason: NotEnrolledReason::NotTargeted, .. }`)
* The bucketing params have changed and the user is no longer in one of the branches
    * (here `status` would match `EnrollmentStatus::NotEnrolled { reason: NotEnrolledReason::NotSelected, .. }`)
* The bucketing params have changed and the user is now assigned to a different branch than before
    * (here `status` would match `EnrollmentStatus::Enrolled { branch: new_branch, .. }` but `new_branch != branch`)
* There was an error while evaluating the updated experiment data
    * (here `status` would match `EnrollmentStatus::Error`)

I think that only the first of these should produce `DisqualifiedReason::NotTargeted`. My understanding is that bucketing changes should just be ignored and they stay in their existing `Enrolled` state and branch, emitting no events. For the error case, perhaps we should introduce a new `DisqualifiedReason::Error` variant to report this?

nimbus/src/enrollment.rs
```diff
@@ -20,28 +24,367 @@ pub enum EnrolledReason {
...
+                ref branch,
+                enrollment_id,
+                ..
+            } => (branch, enrollment_id),
+            EnrollmentStatus::Disqualified {
+                ref branch,
+                enrollment_id,
+                ..
+            } => (branch, enrollment_id),
+            _ => return None, // We were never enrolled anyway, simply the enrollment record from the DB.
```
rfk, line 291: nit: missing word: simply *delete* the enrollment

nimbus/src/enrollment.rs
```diff
@@ -20,28 +24,367 @@ pub enum EnrolledReason {
...
+                    },
+                }
+            }
+            EnrollmentStatus::NotEnrolled { .. } => Self {
+                slug: self.slug.to_string(),
+                status: EnrollmentStatus::NotEnrolled {
+                    reason: NotEnrolledReason::OptOut, // Explicitly set the reason to OptOut.
+                },
+            },
+            _ => {
```
rfk, line 343: A small future-looking nit: would it be worth listing the rest of the states explicitly here rather than using a wildcard `_` match? That way if we add new states in future, the compiler will force us to consider how each new state should handle each of these potential transitions.

nimbus/src/enrollment.rs
```diff
@@ -85,149 +441,232 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
...
+                }
+                (None, None, Some(enrollment)) => enrollment.maybe_garbage_collect(),
+                (None, Some(_), Some(_)) => {
+                    return Err(Error::InternalError(
+                        "New experiment but enrollment already exists.",
+                    ))
+                }
+                (Some(_), None, None) | (Some(_), Some(_), None) => {
+                    return Err(Error::InternalError(
+                        "Every experiment in the DB has an associated enrollment record.",
```
rfk, line 583: micro-nit: when errors occur, it's easier to parse them if they're phrased in terms of what was wrong rather than what was supposed to be right, which in this case would be "Experiment in the db did not have an associated enrollment record")

nimbus/src/enrollment.rs
```diff
@@ -311,55 +758,820 @@ mod tests {
...
+            )?
+            .unwrap();
+        dbg!(&enrollment.status);
+        if let EnrollmentStatus::Enrolled {
+            reason: EnrolledReason::Qualified,
+            enrollment_id: new_enrollment_id,
+            ..
+        } = enrollment.status
+        {
+            assert_eq!(new_enrollment_id, enrollment_id);
```
rfk, line 1047: For completeness, we could also assert that they stayed in the same branch here.

nimbus/src/enrollment.rs
```diff
@@ -311,55 +758,820 @@ mod tests {
...
+        let enrollment = evolver
+            .evolve_enrollment(
+                true,
+                Some(&exp),
+                Some(&exp),
+                Some(&existing_enrollment),
+                &mut events,
+            )?
+            .unwrap();
+        if let EnrollmentStatus::Disqualified {
```
rfk, line 1080: As noted in a comment above, I think the expected behaviour here is that we stay in the same branch rather than becoming disqualified. (As opposed to if the *targetting* had changed, which would disqualify us).

nimbus/src/enrollment.rs
```diff
@@ -311,55 +758,820 @@ mod tests {
...
+            reason: DisqualifiedReason::NotTargeted,
```
rfk, line 1132: As noted in a comment above, I think the expected behaviour here is that we stay in the same branch rather than becoming disqualified.

nimbus/src/enrollment.rs
```diff
@@ -311,55 +758,820 @@ mod tests {
...
+            slug: exp.slug,
+            status: EnrollmentStatus::Disqualified {
+                enrollment_id: Uuid::new_v4(),
+                branch: "control".to_owned(),
+                reason: DisqualifiedReason::NotTargeted,
+            },
+        };
+        let enrollment = existing_enrollment.on_explicit_opt_out(&mut events)?;
+        assert_eq!(enrollment, existing_enrollment);
+        assert!(events.is_empty());
```
rfk, line 1554: As noted in a comment above, I think this should stay in the `Disqualified` state.

nimbus/src/enrollment.rs
```diff
@@ -311,55 +758,820 @@ mod tests {
...
+        Ok(())
+    }
```
rfk, line 1556: A *fantastic* suite of tests, thank you!

For a future iteration, I think it would be neat if we could make these a little more declarative, so that someone like Kate or Tim could review them for correctness without needing to deeply understand the details here. But that's definitely for a followup, not this PR!

rfk (changes requested):
The "request changes" here is specifically for the places where I've noted that the transitions do not match my understanding of intended behaviour. I'm really, really pleased by how well I was able to look at this and identify "this transition does not match my understanding of intended behaviour"!

nimbus/src/enrollment.rs
eoger: This makes sense, I also wonder if we should do something if the branch we were in disappeared.

rfk: Oh yes, good catch! I think `Disqualified` would also make sense in that case.

eoger:
Added the "branch disappears" test case, I think this is ready for a final review pass.

nimbus/src/enrollment.rs
```diff
@@ -20,28 +24,392 @@ pub enum EnrolledReason {
     OptIn,     // Explicit opt-in.
 }
 
+// These are types we use internally for managing non-enrollments.
+#[derive(Deserialize, Serialize, Debug, Clone, Hash, Eq, PartialEq)]
+pub enum NotEnrolledReason {
+    OptOut,      // The user opted-out of experiments before we ever got enrolled to this one.
+    NotSelected, // The evaluator bucketing did not choose us.
+    NotTargeted, // We are not being targeted for this experiment.
+    EnrollmentsClosed, // The experiment enrollment is paused.
```
rfk, line 33: Minor nit: calling this `EnrollmentsPaused` would line up better with the docstring, the name of the field in remote settings, etc.

nimbus/src/enrollment.rs
```diff
@@ -20,28 +23,300 @@ pub enum EnrolledReason {
...
+}
+
+impl ExperimentEnrollment {
+    fn from_new_experiment(
+        is_user_participating: bool,
+        nimbus_id: &Uuid,
+        available_randomization_units: &AvailableRandomizationUnits,
+        app_context: &AppContext,
+        experiment: &Experiment,
+        enrollment_events: &mut Vec<EnrollmentChangeEvent>,
```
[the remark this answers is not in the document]
rfk: I still think this would be clearer as `out_enrollment_events` or something, but it's possible I'm just weird, so I'm not going to press it any further :-)

nimbus/src/enrollment.rs
```diff
@@ -20,28 +24,392 @@ pub enum EnrolledReason {
...
+                experiment,
+            )?;
+            log::debug!(
+                "Experiment '{}' is new - enrollment status is {:?}",
+                &enrollment.slug,
+                &enrollment
+            );
+            if matches!(enrollment.status, EnrollmentStatus::Enrolled { .. }) {
+                enrollment_events.push(enrollment.get_change_event())
+            }
```
rfk, line 90: This cleanup came together really nicely, thanks for trying it out!

nimbus/src/enrollment.rs
```diff
@@ -20,28 +24,392 @@ pub enum EnrolledReason {
...
+                            enrollment_events.push(updated_enrollment.get_change_event());
+                            updated_enrollment
+                        }
+                        EnrollmentStatus::NotEnrolled {
+                            reason: NotEnrolledReason::NotTargeted,
+                        } => {
+                            log::debug!(
+                            "Existing experiment enrollment '{}' is now disqualified (targeting change)",
+                            &self.slug
+                        );
```
rfk, line 206: Some sort of indent error here that wasn't caught by `cargo fmt`?

rfk (approved):
Looks great!

PR #70, opened 2020-12-01
Add normandy_id to available randomization units

piatra:
Currently Firefox only accepts `normandy_id` as randomization unit and drops any other recipe as incompatible.

### DIFF
--- a/nimbus/src/enrollment.rs
+++ b/nimbus/src/enrollment.rs
@@ -513,4 +513,54 @@ mod tests {
         // assert_eq!(enrollments.len(), 1);
         Ok(())
     }
+
+    #[test]
+    fn test_desktop_schema_integration() -> Result<()> {
+        let experiment: serde_json::Value = json!({
+            "schemaVersion": "1.0.0",
+            "slug": "secure-bronze",
+            "endDate": null,
+            "branches":[
+            {"slug": "control", "ratio": 1},
+            {"slug": "treatment","ratio":1}
+            ],
+            "probeSets":[],
+            "startDate":null,
+            "application":"firefox",
+            "bucketConfig":{
+                // Also enroll everyone.
+                "count":10_000,
+                "start":0,
+                "total":10_000,
+                "namespace":"secure-silver",
+                "randomizationUnit":"normandy_id"
+            },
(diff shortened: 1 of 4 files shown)


nimbus/src/enrollment.rs
```diff
@@ -513,4 +513,54 @@ mod tests {
...
+            "isEnrollmentPaused":false,
+            "proposedEnrollment":7,
+            "userFacingDescription":"3rd test experiment.",
+            "id":"secure-bronze",
+            "last_modified":1_602_197_324_372i64
+        });
+        let _ = env_logger::try_init();
+        let tmp_dir = TempDir::new("test_enrollments")?;
+        let db = Database::new(&tmp_dir)?;
+        let normandy_id = Uuid::new_v4();
```
rfk, line 549: nit: I expect this variable is meant to be `nimbus_id` rather than `normandy_id`.

But, I also believe that there have been some significant changes in this file which you'll need to rebase on top of.

nimbus/src/evaluator.rs
```diff
@@ -362,5 +373,20 @@ mod tests {
...
+        let enrollment = evaluate_enrollment(
+            &id,
+            &available_randomization_units,
+            &Default::default(),
+            &experiment3,
+        )
+        .unwrap();
+        assert!(
+            matches!(enrollment.status, EnrollmentStatus::Enrolled { reason: EnrolledReason::Qualified, .. })
+        );
```
rfk, line 390: Please consider also adding a test for the error case, that if the normandy_id is not available then we end up with `EnrollmentStatus::Error`.

rfk (changes requested):
Thanks, this LGTM.

I'm not super excited about adding something that IIUC we hope will be a legacy desktop-only feature, but it makes sense to have it here to work towards desktop integration. @travis79 any objection from you side? We should be able to just totally ignore it in the android component.

Unfortunately I think you'll need to rebase atop a non-trivial refactor of unrollment handling, so I'm requesting changes to ensure I'll come back and look at the result of said rebase.

piatra:
Lets not add this. Like mentioned this is a legacy feature that should be deprecated instead.

PR #73, opened 2020-12-08
Release v0.6.3

rfk:
I tried making this with `cargo release` and it seemed pretty nice, so I've added some instructions to that effect as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ Steps:
 2. Update `CHANGELOG.md` as noted above, and commit your changes.
 3. Run `cargo release --dry-run -vv [major|minor|patch]` and check that the things
    it is proposing to do seem reasonable.
-4. Run `cargo release [major|minor|patch]` to publish the release.
+4. Run `cargo release [major|minor|patch]` to publish the release to github.
+5. Make a PR from your branch to request it be merged to the main branch.
 
 ## Useful Resources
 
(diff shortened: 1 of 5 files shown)


rfk:
It took me a couple of goes to get this right, but it's all build-and-release stuff.

rfk:
Ugh, this is still failing to build as part of https://github.com/mozilla/application-services/pull/3748, due to what I suspect may be a bug in uniffi.

rfk:
Ref https://github.com/mozilla/uniffi-rs/pull/357 for a fix to the uniffi bug which is causing this to fail to build its kotlin bindings over in the appservices release.

rfk:
(To say more explicitly what the next step is: I will make a `v0.6.3` release here with updated uniffi version, once that above PR lands and we publish a new uniffi release)

README.md
```diff
@@ -7,6 +7,20 @@ New and significant features should be listed in the [CHANGELOG.md](./CHANGELOG.
 
 Before issuing a new release, the `Unreleased Changes` section should be renamed to the version that is being released and match with the tagged version, and a new `Unreleased Changes` section added to the top of the document.
 
+## Cutting a release
```
travis79, line 10: Thanks for this!

travis79 (approved):

rfk:
Oh, hrm, it looks like this repo [requires a linear commit history](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/requiring-a-linear-commit-history) so I can't do a merge-commit here, and I've already advanced master from https://github.com/mozilla/nimbus-sdk/pull/75. I'm going to have to do a rebase merge here, meaning that the tagged commit for `v0.6.3` will not be in the history of the main branch :-(

The consequences of this are probably zero in practice, but something to watch out for in future.

PR #74, opened 2020-12-08
Input to choose_branch should not include a dash

piatra:
Is this worth a fix? Came across the issue when trying to generate uuids (for the demo) to match the correct branches.


rfk (approved):
Nice catch. I support fixing it to avoid future confusion, assuming we can get the tests passing without significant change.

PR #75, opened 2020-12-08
Run `cargo update`.

rfk:
This resolves the `cargo audit` issue noted in https://jira.mozilla.com/browse/SDK-144, as well as updating various other things to newer versions. For reference, all I did here was run `cargo update`, run the tests, and commit the result.

### DIFF
--- a/nimbus/Cargo.lock
+++ b/nimbus/Cargo.lock
@@ -20,9 +20,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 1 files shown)


dmose (approved):
Wow, looking at the Cargo.lock, that single `cargo update` added ~16 new (sub-)dependencies.

I think the Rust ecosystem is starting to rival NodeJS.

Anyway, looks good to me...

r=dmose

PR #77, opened 2020-12-10
Implement DB upgrade mechanism

eoger:
The root cause of the [SDK-149](https://jira.mozilla.com/browse/SDK-149) was that we were trying to deserialize variants that did not exist anymore. I think for this specific migration it is okay to throw everything away and re-populate the database, however for the future db upgrades we will have to do proper migrations.

### DIFF
--- a/nimbus/src/persistence.rs
+++ b/nimbus/src/persistence.rs
@@ -15,6 +15,9 @@ use rkv::StoreOptions;
 use std::fs;
 use std::path::Path;
 
+const DB_KEY_DB_VERSION: &str = "db_version";
+const DB_VERSION: u16 = 1; // Increment and implement a DB migration in `maybe_upgrade` when necessary.
+
 // Inspired by Glean - use a feature to choose between the backends.
 // Select the LMDB-powered storage backend when the feature is not activated.
 #[cfg(not(feature = "rkv-safe-mode"))]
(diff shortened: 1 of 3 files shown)


travis79 (approved):
Thanks @eoger!

nimbus/src/persistence.rs
```diff
@@ -15,6 +15,9 @@ use rkv::StoreOptions;
 use std::fs;
 use std::path::Path;
 
+const DB_KEY_DB_VERSION: &str = "db_version";
+const DB_VERSION: u16 = 1; // Increment and implement a DB migration in `maybe_upgrade` when necessary.
```
jhugman, line 19: Suggest we put a comment on the structs and enums that will cause us to increment this number and write a migration.

```rust
// Warning adding/removing to this enum will require incrementing DB_VERSION and a migration.
```

nimbus/src/persistence.rs
```diff
@@ -228,5 +257,61 @@ impl Database {
...
+        assert_eq!(db.get(StoreId::Meta, DB_KEY_DB_VERSION)?, Some(DB_VERSION));
+        assert!(db.collect_all::<String>(StoreId::Enrollments)?.is_empty());
+        assert!(db.collect_all::<String>(StoreId::Experiments)?.is_empty());
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_db_upgrade_unknown_version() -> Result<()> {
+        let path = "test_upgrade_1";
```
jhugman, line 292: Nit: `test_upgrade_2`

jhugman (approved):
A few nits, but nothing to hold up with another round of review. :shipit: 🚢

nimbus/src/persistence.rs
```diff
@@ -147,16 +150,42 @@ impl Database {
...
+        match db_version {
+            Some(DB_VERSION) => return Ok(()),
+            None => {
+                // The "first" version of the database (= no version number) had un-migratable data
+                // for experiments and enrollments, start anew.
+                // XXX: We can most likely remove this behaviour once enough time has passed,
+                // since nimbus wasn't really shipped to production at the time anyway.
+                self.experiment_store.clear(&mut writer)?;
+                self.enrollment_store.clear(&mut writer)?;
+            }
```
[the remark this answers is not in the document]
eoger: Yup, it's not a big deal in terms of performance, but that's why I left the `XXX` comment.

PR #78, opened 2020-12-14
Split update_experiments into two phases

jhugman:
This PR addresses [SDK-142][1].

We split the `update_experiments` method in two to give the [app more control][2] over when experiment changes happen.

Two methods (`set_experiments_locally` and `fetch_experiments`) stash the experiments in a new rkv table called `Updates`.

This table is managed by the methods in `updating.rs`, which provides methods `stash` and a `pop` the updates.

The function `apply_pending_updates` is largely the same as `update_experiments`, but gets its `Experiment`s from the stashed updates. 

`Client::get_experiments` and `http_client::Client::get_experiments` has changed to as to re-use the fault-tolerant deserialization of experiments for both `http_client` and from the new `set_experiments_locally`.

[1]: https://jira.mozilla.com/browse/SDK-142
[2]: https://docs.google.com/document/d/19cPhEV00gmG_BAhvoPDK1eYPtMWVgqmiHxE-xwLSnW0/edit#

### DIFF
--- a/nimbus/examples/experiment.rs
+++ b/nimbus/examples/experiment.rs
@@ -64,8 +64,16 @@ fn main() -> Result<()> {
                 .about("Show all experiments, followed by the enrolled experiments"),
         )
         .subcommand(
+            SubCommand::with_name("fetch-experiments")
+            .about("Fetch experiments from the server. Subsequent calls to apply-pending-experiments will change enrolments."),
+        )
+        .subcommand(
+            SubCommand::with_name("apply-pending-experiments")
+            .about("Updates enrollments with the experiments last fetched from the server with fetch-experiments"),
+        )
+        .subcommand(
             SubCommand::with_name("update-experiments")
-            .about("Updates experiments and enrollments from the server"),
+            .about("Equivalent to fetch-experiments and apply-pending-experiments together"),
         )
         .subcommand(
             SubCommand::with_name("opt-in")
@@ -224,9 +232,19 @@ fn main() -> Result<()> {
                     )
                 });
         }
+        ("fetch-experiments", _) => {
+            println!("======================================");
+            println!("Fetching experiments");
+            nimbus_client.fetch_experiments()?;
+        }
+        ("apply-pending-experiments", _) => {
+            println!("======================================");
+            println!("Applying pending experiments");
+            nimbus_client.apply_pending_experiments()?;
+        }
         ("update-experiments", _) => {
             println!("======================================");
-            println!("Updating experiments");
+            println!("Fetching and applying experiments");
             nimbus_client.fetch_experiments()?;
             nimbus_client.apply_pending_experiments()?;
         }
--- a/nimbus/examples/experiment.rs
+++ b/nimbus/examples/experiment.rs
@@ -199,7 +199,8 @@ fn main() -> Result<()> {
     let mut nimbus_client = NimbusClient::new(context, "", Some(config), aru)?;
 
     // Explicitly update experiments at least once for init purposes
-    nimbus_client.update_experiments()?;
+    nimbus_client.fetch_experiments()?;
+    nimbus_client.apply_pending_experiments()?;
 
     // We match against the subcommands
     match matches.subcommand() {
@@ -226,7 +227,8 @@ fn main() -> Result<()> {
         ("update-experiments", _) => {
             println!("======================================");
             println!("Updating experiments");
-            nimbus_client.update_experiments()?;
+            nimbus_client.fetch_experiments()?;
+            nimbus_client.apply_pending_experiments()?;
         }
         ("opt-in", Some(matches)) => {
             println!("======================================");
--- a/nimbus/src/client/http_client.rs
+++ b/nimbus/src/client/http_client.rs
@@ -17,7 +17,8 @@ use std::time::{Duration, Instant};
 
 use crate::config::RemoteSettingsConfig;
 use crate::error::{Error, Result};
-use crate::{Experiment, SettingsClient, SCHEMA_VERSION};
+use crate::{Experiment, Experiments, SettingsClient, SCHEMA_VERSION};
+use std::convert::TryFrom;
 use url::Url;
 use viaduct::{status_codes, Request, Response};
 
@@ -109,7 +110,7 @@ impl SettingsClient for Client {
         unimplemented!();
     }
 
-    fn get_experiments(&mut self) -> Result<Vec<Experiment>> {
+    fn fetch_experiments(&mut self) -> Result<Experiments> {
         let path = format!(
             "buckets/{}/collections/{}/records",
             &self.bucket_name, &self.collection_name
@@ -120,7 +121,15 @@ impl SettingsClient for Client {
         // to allow us to deserialize each experiment individually,
         // omitting any malformed experiments
         let resp = self.make_request(req)?.json::<serde_json::Value>()?;
-        let data = resp.get("data").ok_or(Error::InvalidExperimentResponse)?;
+        let res = Experiments::try_from(resp)?;
+        Ok(res)
+    }
+}
+
+impl TryFrom<serde_json::Value> for Experiments {
+    type Error = crate::error::Error;
+    fn try_from(value: serde_json::Value) -> Result<Self> {
+        let data = value.get("data").ok_or(Error::InvalidExperimentResponse)?;
         let mut res = Vec::new();
         for exp in data.as_array().ok_or(Error::InvalidExperimentResponse)? {
             // Validate the schema major version matches the supported version
@@ -157,7 +166,7 @@ impl SettingsClient for Client {
                 }
             }
         }
-        Ok(res)
+        Ok(Self::new(res))
     }
 }
 
@@ -238,7 +247,7 @@ mod tests {
     }
 
     #[test]
-    fn test_get_experiments_from_schema() {
+    fn test_fetch_experiments_from_schema() {
         viaduct_reqwest::use_reqwest_backend();
         // There are two experiments defined here, one has a "newer" schema version
         // in order to test filtering of unsupported schema versions.
@@ -256,7 +265,8 @@ mod tests {
             collection_name: "messaging-experiments".to_string(),
         };
         let mut http_client = Client::new(config).unwrap();
-        let resp = http_client.get_experiments().unwrap();
+        let resp = http_client.fetch_experiments().unwrap();
+        let resp = resp.get_experiments();
         m.expect(1).assert();
         assert_eq!(resp.len(), 1);
         let exp = &resp[0];
@@ -317,8 +327,8 @@ mod tests {
             collection_name: "messaging-experiments".to_string(),
         };
         let mut http_client = Client::new(config).unwrap();
-        assert!(http_client.get_experiments().is_ok());
-        let second_request = http_client.get_experiments();
+        assert!(http_client.fetch_experiments().is_ok());
+        let second_request = http_client.fetch_experiments();
         assert!(matches!(second_request, Err(Error::BackoffError(_))));
         m.expect(1).assert();
     }
@@ -340,8 +350,8 @@ mod tests {
             collection_name: "messaging-experiments".to_string(),
         };
         let mut http_client = Client::new(config).unwrap();
-        assert!(http_client.get_experiments().is_err());
-        let second_request = http_client.get_experiments();
+        assert!(http_client.fetch_experiments().is_err());
+        let second_request = http_client.fetch_experiments();
         assert!(matches!(second_request, Err(Error::BackoffError(_))));
         m.expect(1).assert();
     }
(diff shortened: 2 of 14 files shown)


nimbus/src/persistence.rs
```diff
@@ -152,11 +154,13 @@ impl Database {
         let meta_store = rkv.open_single("meta", StoreOptions::create())?;
         let experiment_store = rkv.open_single("experiments", StoreOptions::create())?;
         let enrollment_store = rkv.open_single("enrollments", StoreOptions::create())?;
+        let updates_store = rkv.open_single("updates", StoreOptions::create())?;
```
jhugman, line 157: We're adding another store here, but because it's cleared most of the time, I don't think it needs to alter the schema or have a migration added.

nimbus/tests/test_updates.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[cfg(feature = "rkv-safe-mode")]
+    #[test]
+    fn test_startup_behavior() -> Result<()> {
+        let mut client = new_client("test_startup_behavior")?;
+        startup(&mut client, true)?;
+
+        let experiments = client.get_all_experiments()?;
+        assert_eq!(experiments.len(), 2);
+        assert_eq!(experiments[0].slug, "secure-gold");
+        assert_eq!(experiments[1].slug, "startup-gold");
```
jhugman, line 185: I am puzzled why this is not in the order given.

nimbus/src/updating.rs
```diff
@@ -0,0 +1,71 @@
...
+pub fn pop_pending_updates(db: &Database, writer: &mut Writer) -> Result<Option<Experiments>> {
+    let store = db.get_store(StoreId::Updates);
+    let experiments = store.get::<Experiments>(writer, KEY_PENDING_UPDATES)?;
+
+    if let Some(_) = experiments {
+        store.clear(writer)?;
+    }
+
+    Ok(experiments)
+}
```
jhugman, line 31: Use an `Option` here to distinguish between "no pending updates", and "updates of an empty list", i.e. all experiments ended.

nimbus/src/nimbus.idl
```diff
@@ -87,12 +87,31 @@ interface NimbusClient {
     [Throws=Error]
     sequence<EnrollmentChangeEvent> set_global_user_participation(boolean opt_in);
 
-    // Updates the list of experiments from the server. After calling this, the
-    // list of active experiments might change (there might be new experiments,
-    // or old experiments might have expired)
+    // Updates the list of experiments from the server.
+    // This method is deprecated, in favour of calling `fetch_experiments()` and then
```
mhammond, line 91: I guess more a question for @travis79, but do we really need to deprecate (ie, can we just also make the somewhat trivial PR for a-c?) If we do, it's probably worth a `#[deprecated = "blah"]` decoration on the implementation?

nimbus/src/persistence.rs
mhammond: IIUC, it would be safe to unconditionally delete all records from this store in those scenarios, whereas doing nothing carries some risk?

nimbus/src/updating.rs
```diff
@@ -0,0 +1,71 @@
...
+//! This module implements the primitive functions to implement
+//! safe updating from the server.
+
+use crate::error::Result;
+use crate::persistence::{Database, StoreId, Writer};
+use crate::Experiments;
+
+const KEY_PENDING_UPDATES: &str = "pending-experiment-updates";
+
+pub fn stash_pending_updates(db: &Database, experiments: Experiments) -> Result<()> {
```
mhammond, line 14: nit: not a huge fan of "stash" here when no other functions which write use it - and the term makes me think "oh, it's doing something write-like, but fancy?" - but it turns out it's just writing them. Contrast with `pop, which I think *is* fine - that name implies "read and remove" - to me stash implies "write and ???<profit?>" :)

OTOH, most other functions which do write don't explicitly use "write" or "save" etc in their name either - but it's worth considering.

nimbus/src/nimbus.idl
```diff
@@ -87,12 +87,31 @@ interface NimbusClient {
...
     [Throws=Error]
     sequence<EnrollmentChangeEvent> update_experiments();
 
+    // Fetches the list of experiments from the server. This does not affect the list
+    // of active experiments or experiment enrolment.
+    // Fetched experiments are not applied until `apply_pending_updates()` is called.
+    [Throws=Error]
+    void fetch_experiments();
+
+    // Apply the updated experiments from the last fetch.
```
mhammond, line 102: Naming nits, and I apologize for not bringing these up in your doc. I'm also not going to die on this hill :) But:

In `apply_pending_updates()`, the `pending` and `updates` part are somewhat redundant. I wonder if `apply_pending_experiments()` is a better name? ie, you "fetch experiments", then you have "pending experiments" which you can apply.

nimbus/src/nimbus.idl
```diff
@@ -87,12 +87,31 @@ interface NimbusClient {
...
+    // (there might be new experiments, or old experiments might have expired).
+    [Throws=Error]
+    sequence<EnrollmentChangeEvent> apply_pending_updates();
+
+    // A convenience method for apps to set the experiments from a local source
+    // for either testing, or before the first fetch has finished.
+    // 
+    // Experiments set with this method are not applied until `apply_pending_updates()` is called.
+    [Throws=Error]
+    void set_experiments_locally(string experiments_json);
```
mhammond, line 113: And here, how about something that's a little clearer about what it's actually doing - `stage_pending_experiments()` springs to mind - it's a littler clearer that this isn't anything new conceptually than what's already described above.

nimbus/tests/test_updates.rs
mhammond: TIL that rkv iterators "returns the (key, value) tuples in order", which seems to explain it.

mhammond (changes requested):
This looks great! This turned into many more comments and nits than I was expecting, but feel free to push back if you feel strongly about any of my comments (although I feel somewhat strongly about the TryInfo ;). I wouldn't mind another look though.

rfk (commented):
Please also add a changelog entry for this, for completeness.

rfk:
Cross-linking https://github.com/mozilla/nimbus-sdk/pull/80 for context, since it talks a bit about pending updates in the context of opting out of telemetry.

nimbus/src/client/null_client.rs
```diff
@@ -36,7 +36,8 @@ fn test_null_client() -> Result<()> {
     let tmp_dir = TempDir::new("test_null_client-test_null")?;
 
     let aru = Default::default();
-    let client = NimbusClient::new(Default::default(), tmp_dir.path(), None, aru)?;
+    let mut client = NimbusClient::new(Default::default(), tmp_dir.path(), None, aru)?;
+    client.update_experiments()?;
```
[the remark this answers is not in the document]
jhugman: It's testing that the `null_client` returns an empty list of experiments. There are more direct ways of testing this, of course; and I should really re-write this test to exercise the `fetch_experiments` method directly.

(`update_experiments` calls `fetch_experiments`, but `set_local_experiments` does not)

nimbus/src/updating.rs
jhugman: Renamed these to be more explicit.

nimbus/src/nimbus.idl
jhugman: I like your suggestion of removing `updates`, and settled for `experiments` everywhere.

nimbus/src/nimbus.idl
jhugman: For some reason, I don't like `stage` here. We're also using `fetchExperiments()`, and `applyPendingExperiments()`, which would suggest we could use `writeExperiments(string)` or `setPendingExperiments(string)`.

(Notice that we're not accumulating experiments to be applied; one `fetchExperiments` will overwrite another.

If this were a Swift API, I think I'd call this: 

```swift
func setExperiments(from string: String)

// use like: 
nimbusClient.setExperiment(from: """{
    "data": [
         …elided…
     ]
} """)
```

But then perhaps we this method should re-calculate enrollments immediately, without the app having to call `apply` itself.

nimbus/src/nimbus.idl
travis79: Yeah, with the lack of consumers I think we can be safe to move fast and skip the deprecation.

jhugman: This will cause unavoidable build breakage for android-components.

nimbus/src/nimbus.idl
jhugman: I've left this as is; I like the API called `set_experiments_locally()`, I'm not wild about `set_pending_experiments()` and dislike `stage` anything.

jhugman (commented):
I think I'm ready for a re-review on this now @mhammond !

/me waves.

nimbus/examples/experiment.rs
```diff
@@ -226,7 +227,8 @@ fn main() -> Result<()> {
         ("update-experiments", _) => {
             println!("======================================");
             println!("Updating experiments");
-            nimbus_client.update_experiments()?;
+            nimbus_client.fetch_experiments()?;
+            nimbus_client.apply_pending_experiments()?;
```
mhammond, line 231: I think it would be idea if we split this into 2 commands (ie, `fetch-experiments` and `apply-...`) just so this tool ends up being able to reproduce the state real apps find themselves in). I'm not too bothered if you want to do this later though because for this to make sense we probably want to print the state of the DB (ie, so the user could see there are fetched-but-not-applied experiments pending) and I'm not going to block you on that kind of change (but would welcome it ;) At a minimum though, I think changing the `println` above is worthwhile to use the same terminology (eg, `Fetching and applying experiments` or similar) just to set expectations correctly.

nimbus/src/client/http_client.rs
```diff
@@ -119,46 +119,54 @@ impl SettingsClient for Client {
         // We first encode the response into a `serde_json::Value`
```
mhammond, line 119: This comment block probably now belongs in `parse_experiments`?

nimbus/src/client/http_client.rs
```diff
@@ -119,46 +119,54 @@ impl SettingsClient for Client {
...
-            let schema_maj_version = exp_schema_version.split('.').next().unwrap_or("");
-            // While "0" is a valid schema version, we have already passed that so reserving zero as
-            // a special value here in order to avoid a panic, and just ignore the experiment.
-            let schema_version: u32 = schema_maj_version.parse().unwrap_or(0);
-            if schema_version != SCHEMA_VERSION {
-                log::info!(
+        let resp = self.make_request(req)?;
+        let string = resp.text();
+
+        parse_experiments(&string)
```
mhammond, line 125: nit: I'd be inclined to fold the line above into this one

nimbus/src/sampling.rs
```diff
@@ -143,7 +143,7 @@ fn is_hash_in_bucket(
 /// # Errors
 /// returns an error if the fraction not within the 0-1 range
 fn fraction_to_key(fraction: f64) -> Result<String> {
-    if fraction < 0.0 || fraction > 1.0 {
+    if !(0.0..=1.0).contains(&fraction) {
```
mhammond, line 146: heh - this is cute and all, but not sure the intent is as clear as the "obvious" version, but 🤷  :)

nimbus/src/updating.rs
mhammond: Might be worth adding this note as a comment? For bonus points, maybe mention why the caller would care? A quick look shows this is a micro-optimization, so it's probably worth calling out that there's no logical different between `None` and `Some(vec![])`?

nimbus/src/updating.rs
```diff
@@ -0,0 +1,72 @@
...
+    let experiments = store.get::<Vec<Experiment>>(writer, KEY_PENDING_UPDATES)?;
+
+    if experiments.is_some() {
+        store.clear(writer)?;
+    }
+
+    Ok(experiments)
+}
+
+#[cfg(feature = "rkv-safe-mode")]
```
mhammond, line 36: I guess this test also crashes rkv for no good reason? If so, please add a comment indicating this (and if not, then please remove it :)

mhammond (approved):
👍

nimbus/src/sampling.rs
jhugman: TBH, I was trying to fix the clippy errors. This has nothing to do with the rest of this patch.

nimbus/examples/experiment.rs
jhugman: I've added the two additional commands; however, I'm going to file an issue to consider removing the [initial update][1] and adding a subcommand for `set_local_experiments` (to set a file).

[1]: https://github.com/mozilla/nimbus-sdk/pull/78/files#diff-ab0d747cc0a25ee15aaa963c028e831d3bac3cff22997e0b4784ca4993cfc5d6R202-R203

jhugman: Filed [SDK-162][1] for adding commands for testing startup.

[1]: https://jira.mozilla.com/browse/SDK-162

PR #80, opened 2020-12-15
Add `reset_telemetry_identifiers` method

rfk:
Fixes [SDK-136](https://jira.mozilla.com/browse/SDK-136).

When the user disables telemetry in the consuming application, we need
to reset any unique identifiers used internally by the SDK. This guards
against accidentally correlating metrics submitted before the opt-out
with metrics that the user may opt-in to submitting again in future.

Consumers are expected to call `on_telemetry_disabled` when the user
disables telemetry in order to trigger resetting of the identifiers.

This needs tests before it's ready for review, but I wanted to push it early for feedback in light of https://github.com/mozilla/nimbus-sdk/pull/78. IIUC, the work in https://github.com/mozilla/nimbus-sdk/pull/78 is based on the notion that we should not change experiment enrollment state as an incidental side-effect of other operations in the SDK, because doing so could cause weird UX changes or inconsistencies in the app. Similarly, I think it would be weird for the user to click "disable telemetry" and then suddenly some unrelated part of the application UX changes.

~~So, what I've done here is add an `on_telemetry_disabled` method which resets the unique identifiers used by the SDK, but *leaves their existing enrollments intact*.~~

~~I think it would be reasonable to wipe any existing enrollments in the next call to `apply_pending_updates`, but we should defer that until after #78 lands so we can re-use its "pending update" infrastructure.~~

[**UPDATE:** the approach in this PR has changed a bit, see comments below for details]

@travis79 @mhammond @jhugman thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,20 @@
 # Unreleased changes
+
 ## What's New
- - Split up `NimbusClient.update_experiments()` into a slow `NimbusClient.fetch_experiments()` and a fast `NimbusClient.apply_pending_experiments()` to help apps manage concurrency and mutable state.
- - Add `set_local_experiments(string)`, to help apps, build tooling for tests, and help during startup on first time run.
+
+- Split up `NimbusClient.update_experiments()` into a slow `NimbusClient.fetch_experiments()`
+  and a fast `NimbusClient.apply_pending_experiments()` to help apps manage concurrency and
+  mutable state.
+- Add `set_local_experiments(string)`, to help apps, build tooling for tests, and help during
+  startup on first time run.
+- A new `reset_telemetry_identifiers` method has been added; consumers should arrange to call
+  this method if the user opts out of telemetry at the application level, in order to avoid
+  accidental tracking of users who disable then re-enable telemetry.
 
 ## ⚠️ Breaking changes ⚠️
- - `NimbusClient.updateExperiments()` is removed.
- - Renamed `InvalidExperimentResponse` error to `InvalidExperimentFormat`.
+- `NimbusClient.updateExperiments()` is removed.
+- Renamed `InvalidExperimentResponse` error to `InvalidExperimentFormat`.
+
 # 0.6.4 (_2020-12-16_)
 
 ## What's New
--- a/nimbus/src/enrollment.rs
+++ b/nimbus/src/enrollment.rs
@@ -149,36 +149,20 @@ impl ExperimentEnrollment {
                     updated_enrollment
                 }
             }
-            EnrollmentStatus::Enrolled {
-                ref branch,
-                enrollment_id,
-                ..
-            } => {
+            EnrollmentStatus::Enrolled { ref branch, .. } => {
                 if !is_user_participating {
                     log::debug!(
                         "Existing experiment enrollment '{}' is now disqualified (global opt-out)",
                         &self.slug
                     );
-                    let updated_enrollment = Self {
-                        slug: self.slug.clone(),
-                        status: EnrollmentStatus::Disqualified {
-                            reason: DisqualifiedReason::OptOut,
-                            enrollment_id,
-                            branch: branch.clone(),
-                        },
-                    };
+                    let updated_enrollment =
+                        self.disqualify_from_enrolled(DisqualifiedReason::OptOut);
                     out_enrollment_events.push(updated_enrollment.get_change_event());
                     updated_enrollment
                 } else if !updated_experiment.has_branch(branch) {
                     // The branch we were in disappeared!
-                    let updated_enrollment = Self {
-                        slug: self.slug.clone(),
-                        status: EnrollmentStatus::Disqualified {
-                            reason: DisqualifiedReason::Error,
-                            enrollment_id,
-                            branch: branch.clone(),
-                        },
-                    };
+                    let updated_enrollment =
+                        self.disqualify_from_enrolled(DisqualifiedReason::Error);
                     out_enrollment_events.push(updated_enrollment.get_change_event());
                     updated_enrollment
                 } else {
@@ -190,14 +174,8 @@ impl ExperimentEnrollment {
                     )?;
                     match evaluated_enrollment.status {
                         EnrollmentStatus::Error { .. } => {
-                            let updated_enrollment = Self {
-                                slug: self.slug.clone(),
-                                status: EnrollmentStatus::Disqualified {
-                                    reason: DisqualifiedReason::Error,
-                                    enrollment_id,
-                                    branch: branch.clone(),
-                                },
-                            };
+                            let updated_enrollment =
+                                self.disqualify_from_enrolled(DisqualifiedReason::Error);
                             out_enrollment_events.push(updated_enrollment.get_change_event());
                             updated_enrollment
                         }
@@ -205,14 +183,8 @@ impl ExperimentEnrollment {
                             reason: NotEnrolledReason::NotTargeted,
                         } => {
                             log::debug!("Existing experiment enrollment '{}' is now disqualified (targeting change)", &self.slug);
-                            let updated_enrollment = Self {
-                                slug: self.slug.clone(),
-                                status: EnrollmentStatus::Disqualified {
-                                    reason: DisqualifiedReason::NotTargeted,
-                                    enrollment_id,
-                                    branch: branch.clone(),
-                                },
-                            };
+                            let updated_enrollment =
+                                self.disqualify_from_enrolled(DisqualifiedReason::NotTargeted);
                             out_enrollment_events.push(updated_enrollment.get_change_event());
                             updated_enrollment
                         }
@@ -296,19 +268,8 @@ impl ExperimentEnrollment {
         out_enrollment_events: &mut Vec<EnrollmentChangeEvent>,
     ) -> Result<Self> {
         Ok(match self.status {
-            EnrollmentStatus::Enrolled {
-                enrollment_id,
-                ref branch,
-                ..
-            } => {
-                let enrollment = Self {
-                    slug: self.slug.to_string(),
-                    status: EnrollmentStatus::Disqualified {
-                        reason: DisqualifiedReason::OptOut,
-                        branch: branch.to_owned(),
-                        enrollment_id,
-                    },
-                };
+            EnrollmentStatus::Enrolled { .. } => {
+                let enrollment = self.disqualify_from_enrolled(DisqualifiedReason::OptOut);
                 out_enrollment_events.push(enrollment.get_change_event());
                 enrollment
             }
@@ -325,6 +286,37 @@ impl ExperimentEnrollment {
                 self.clone()
             }
         })
+    }
+
+    /// Reset identifiers in response to application-level telemetry reset.
+    ///
+    /// Each experiment enrollment record contains a unique `enrollment_id`. When the user
+    /// resets their application-level telemetry, we reset each such id to a special nil value,
+    /// creating a clean break between data sent before the reset and any data that might be
+    /// submitted about these enrollments in future.
+    ///
+    /// We also move any enrolled experiments to the "disqualified" state, since their further
+    /// partipation would submit partial data that could skew analysis.
+    ///
+    fn reset_telemetry_identifiers(
+        &self,
+        out_enrollment_events: &mut Vec<EnrollmentChangeEvent>,
+    ) -> Self {
+        let updated = match self.status {
+            EnrollmentStatus::Enrolled { .. } => {
+                let disqualified = self.disqualify_from_enrolled(DisqualifiedReason::OptOut);
+                out_enrollment_events.push(disqualified.get_change_event());
+                disqualified
+            }
+            EnrollmentStatus::NotEnrolled { .. }
+            | EnrollmentStatus::Disqualified { .. }
+            | EnrollmentStatus::WasEnrolled { .. }
+            | EnrollmentStatus::Error { .. } => self.clone(),
+        };
+        ExperimentEnrollment {
+            status: updated.status.clone_with_nil_enrollment_id(),
+            ..updated
+        }
     }
 
     /// Garbage collect old experiments we've kept a WasEnrolled enrollment from.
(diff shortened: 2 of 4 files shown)


nimbus/src/lib.rs
```diff
@@ -138,6 +138,42 @@ impl NimbusClient {
...
+        let store = db.get_store(StoreId::Meta);
+        Ok(
+            if let Some(_) = store.get::<String>(&writer, DB_KEY_NIMBUS_ID)? {
+                // Each enrollment state includes a unique `enrollment_id`.
+                // TODO: Ideally, we would wipe existing experiment enrollments at the next
+                // available opportunity. We don't do it immediately because we don't want
+                // to trigger weird changes in application UI.
+                // We could add to the "pending changes" infra from https://github.com/mozilla/nimbus-sdk/pull/78
+                // once it lands.
+                enrollment::on_telemetry_disabled(db, &mut writer)?;
```
rfk, line 169: If we do end up enqueuing a "wipe enrollments" flag to be applied at the next startup, we may not need to reset the enrollment identifiers at all.

mhammond:
>  Similarly, I think it would be weird for the user to click "disable telemetry" and then suddenly some unrelated part of the application UX changes.
...
> I think it would be reasonable to wipe any existing enrollments in the next call to `apply_pending_updates`, but we should defer that until after #78 lands so we can re-use its "pending update" infrastructure.

This isn't a comment on the PR, but just on the possible approach you outline here. I believe the intention is that `apply_pending_updates()` would be called somewhat regularly, so it's still likely that there will be a perception that clicking "disable telemetry" causes some unrelated part of the application UX to change.

IMO, it might make sense to leave existing enrollments alone entirely - let them end when they would have ended anyway. Obviously new enrollments would not happen for such users, but it seems like there will be less surprises for users to have "disable telemetry" and "app UX changes" being as decoupled as possible.

As noted above, this is more a policy decision than a comment on this PR (and indeed, the outcome of that decision would impact only the comments in this PR) but I thought it worth raising early. I'll look at the PR in detail soon.

nimbus/src/enrollment.rs
```diff
@@ -427,12 +435,34 @@ impl EnrollmentStatus {
...
+                ..
+            }
+            | EnrollmentStatus::WasEnrolled {
+                ref mut enrollment_id,
+                ..
+            } => *enrollment_id = Uuid::new_v4(),
+            EnrollmentStatus::NotEnrolled { .. } | EnrollmentStatus::Error { .. } => (),
+        };
+        updated
+    }
```
travis79, line 465: Do we want to keep the current enrollment ids?  Since we are only keeping the experiment enrollments to prevent some changes from running away into the app, and we aren't sending any of this in telemetry, I don't know that we need to worry about rotating the enrollment_ids here.  If the client re-enabled telemetry, some of these enrollments might get weird if the randomization unit changes and they are no longer in a targeted bucket.

nimbus/src/enrollment.rs
```diff
@@ -690,6 +720,30 @@ pub fn opt_out(db: &Database, experiment_slug: &str) -> Result<Vec<EnrollmentCha
...
+/// Reset unique identifiers in response to application-level telemetry opt-out.
+///
+/// Each experiment enrollment record contains a unique `enrollment_id`. When the user
+/// opts out of application-level telemetry, we reset each such id to a new random value,
+/// creating a clean break between data sent before the opt-out and any data they might
+/// send in future.
+///
+/// Note that we do *not* change their experiment enrollments, because we don't want
+/// the act of disabling telemetry to cascade into other unrelated changes in the app
+/// as the user moves out of active experiment branches.
```
travis79, line 732: This raises an interesting concern if the user re-enables telemetry with a new randomization unit that no longer qualifies the client bucket for the experiment. I'm not quite sure what to do with the existing enrollments, and how do we know if the user enables telemetry again?

nimbus/src/enrollment.rs
rfk: > Do we want to keep the current enrollment ids?

I don't think we should. The case I'm worried about here is something like:

1. User gets enrolled in an experiment, emits a telemetry `Enrolled` event with `enrollment_id=AAA`.
2. User disables telemetry.
3. User re-enables telemetry, remaining enrolled with the same enrollment id.
4. The experiment ends, emitting a telemetry `Unenrolled` event with `enrollment_id=AAA`.

We could use the events from (1) and (4) to link the user's telemetry from before disablement with their telemetry after re-enablement.

nimbus/src/enrollment.rs
rfk: Indeed. I'm wondering if we can move any existing enrolled experiments to `Disqualified` state to prevent weird client-side behaviours or data analysis artifacts if the user re-enables telemetry.

nimbus/src/enrollment.rs
rfk: (OTOH, in theory we're supposed to have deleted the user's telemetry from before disablement, so maybe that's not as important as it might seem...?)

rfk: As a point of comparison, [the desktop implementation](https://searchfox.org/mozilla-central/rev/8883276967d39918e2ce64e873afdd432fb406ca/toolkit/components/normandy/lib/PreferenceExperiments.jsm#387) sets enrollment ids to a canary value of `__NO_ENROLLMENT_ID__`.

rfk:
On reflection, over-optimising for the upcoming "apply pending updates" case seems like a mistake. What we're doing here is very similar to the current `set_global_user_participation` behaviour, which IIUC we have not yet figured out it should interact with a `apply_pending_updates` method. Instead, it just moves the current enrollment state directly to `Disqualified`.

Let's do the same here: reset the identifiers, move currently-enrolled experiments to `Disqualified` to avoid complicating things if the user re-enables telemetry, and iterate the interaction with `apply_pending_updates` if necessary in future.

I'll push an update accordingly.

nimbus/src/enrollment.rs
```diff
@@ -149,36 +149,19 @@ impl ExperimentEnrollment {
...
-                    let updated_enrollment = Self {
-                        slug: self.slug.clone(),
-                        status: EnrollmentStatus::Disqualified {
-                            reason: DisqualifiedReason::OptOut,
-                            enrollment_id,
-                            branch: branch.clone(),
-                        },
-                    };
+                    let updated_enrollment =
+                        self.disqualify_if_enrolled(DisqualifiedReason::OptOut);
```
rfk, line 159: We do this disqualification pattern in enough places that it seemed like a helper method would be useful.

rfk:
I think this is ready for review with the modified approach.

nimbus/src/enrollment.rs
```diff
@@ -327,6 +287,28 @@ impl ExperimentEnrollment {
...
+    /// opts out of application-level telemetry, we reset each such id to a special nil value,
+    /// creating a clean break between data sent before the opt-out and any data that might be
+    /// submitted about these enrollments in future.
+    ///
+    /// We also move any enrolled experiments to the "disqualified" state, since their
+    /// further partipation in the experiment may skew the data if they later re-enable
+    /// telemetry.
+    ///
+    /// Note that this does not return any change events; since the user has opted out
+    /// of telemetry, we wouldn't be able to submit said events anyway.
```
rfk, line 302: Hrm, according to https://mana.mozilla.org/wiki/pages/viewpage.action?spaceKey=FJT&title=Nimbus+Engineering#NimbusEngineering-Events we're supposed to emit an `Unenrollment` event when telemetry is disabled. My impression was that, since the user has disabled telemetry, we would not try to send any further events. @travis79 do you have any insight into the correct behaviour here?

nimbus/src/enrollment.rs
travis79: You are right, if we kept the same enrollments and ids when re-enabling telemetry, if we sent out any events, it would probably look like a new user (because the telemetry id was rotated) that had an unenrollment (or disqualification) event without a matching enrollment event. So, resetting the ids to a canary value I would definitely approve of (we use a `c0ffee` uuid in Telemetry for this).  

I would really prefer a way to just delete all the enrollment info just like we delete all the telemetry associated with a client, but I do understand the concern about what that might mean to the application if we can't guarantee the same result to multiple calls of getBranch, etc. Maybe this is a good question to bring up with the product team(s) to see what sort of behavior they would expect in this case? I don't know that this really needs to be addressed immediately, rather a good to think about as a possible iteration.

nimbus/src/enrollment.rs
travis79: Moving them to disqualified seems reasonable to me, along with resetting the enrollment_id to a canary value.

nimbus/src/enrollment.rs
travis79: If telemetry is disabled, it won't make it out. Glean drops all current telemetry and sends a deletion_request ping.  What we might want to do in this case is include the enrollment_id in the deletion_request ping as a sort of unenrollment event.  I'm not certain what we can or would do with the ids (or the event for that matter) since the deletion_request ping will eventually erase all unaggregated data for that user anyway.

travis79 (changes requested):
One more thing that occurred to me, if we are holding on to any other randomization units, we probably want to reset those also (client_id, normandy_id, etc.)

rfk:
@travis79 to clarify, is the "request changes" here specifically about dealing with other randomization units?

travis79:
> @travis79 to clarify, is the "request changes" here specifically about dealing with other randomization units?

Yes, I probably meant to click "comment" but this was my main concern.

rfk:
From a quick look, yeah, you're right that `self.available_randomization_units` holds a potential reference to the telemetry client_id and the application will need to provide a new value. I'm in two minds about the right API here:

* We could accept updated values as an argument, like `on_telemetry_disabled(updated_randomization_units)`
* We could ask the application to re-create the `NimbusClient` instance and pass it the updated values.

I'll play around with it and see if I can find something I like.

rfk:
> We could accept updated values as an argument, like on_telemetry_disabled(updated_randomization_units)

This isn't going to work, because we may not know the new identifiers until the time at which telemetry is re-enabled.

rfk:
Well this was a bit of a can of worms, but I think I have a sensible approach.

I started out by making a mock PR to Fenix, which you can find at https://github.com/mozilla-mobile/fenix/pull/17087. In this context  I think the name `resetTelemetryIdentifiers` is a better fit than `onTelemetryDisabled`, because the app is using it more as an imperative command than as a notification callback. A future version of this code might have to construct a new `AvailableRandomizationUnits` when callings this method, but for now I took the same approach as we do at initialization time and papered over it in the android-components later, ref https://github.com/travis79/android-components/pull/6.

Here in the Rust code, we now have a `reset_telemetry_identifiers` method that:

* Accepts any new `AvailableRandomizationUnits` from the application.
* Resets the internal `nimbus_id`.
* Resets any enrollment ids to a canary value as [the nil UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier#Nil_UUID).
* Moves any enrolled experiments to `Disqualified` state.
* Returns change events for the above.

I expect that the application will not end up submitting the change events, because telemetry has been disabled! But emitting them seems more in keeping with the shape of the rest of the API.

travis79 (commented):
This approach looks good to me, but I still don't understand passing in "new" identifiers to the reset telemetry function as I don't think we would have them until it was re-enabled.  This makes me worry that they won't call this until telemetry is re-enabled when they have that information to pass in. I don't imagine that this would cause any problems to call it when telemetry is disabled with `null` and call it again when telemetry is re-enabled and pass in the new randomization units?  Seems a little awkward but not terrible.  

Honestly, I don't see the case of people turning telemetry off and then back on a whole lot in the wild, so maybe that is chasing an edge case and it's okay to just reset everything and count on getting the new identifiers on the next launch and call to init.

rfk:
> This approach looks good to me, but I still don't understand passing in "new" identifiers to the reset telemetry function
> as I don't think we would have them until it was re-enabled

It's not obvious because we don't currently pass in any identifiers in practice, but this also lets you explicitly pass in `client_id = None` at the point of disabling telemetry. I suppose we could just do that internally without requiring the caller, but it's not how much Nimbus is supposed to know about the semantics of those externally-provided identifiers.

mhammond:
>  it's okay to just reset everything and count on getting the new identifiers on the next launch and call to init.

FWIW, this sounds like exactly the right thing to do. I guess we'd need some decision maker to sign off on that though?

travis79 (approved):
Doh, I didn't think about passing in the empty version of the randomization units. I think that this is fine since we can pass "empty" identifiers and I don't feel the need to bikeshed too much on this. We can always address any ergonomic issues that might arise around this in later iterations.

CHANGELOG.md
```diff
@@ -1,11 +1,19 @@
...
- - Split up `NimbusClient.update_experiments()` into a slow `NimbusClient.fetch_experiments()` and a fast `NimbusClient.apply_pending_experiments()` to help apps manage concurrency and mutable state.
- - Add `set_local_experiments(string)`, to help apps, build tooling for tests, and help during startup on first time run.
+- Split up `NimbusClient.update_experiments()` into a slow `NimbusClient.fetch_experiments()`
+  and a fast `NimbusClient.apply_pending_experiments()` to help apps manage concurrency and
+  mutable state.
+- Add `set_local_experiments(string)`, to help apps, build tooling for tests, and help during
+  startup on first time run.
+- A new `reset_telemetry_identifiers` method has been added; consumers should arrange to call
+  this method if the user opts out of telemetry at the application level, in order to avoid
+  accidental tracking of users who disable then re-enable telemetry.
```
rfk, line 11: (The changes here are formatting consistency issues that I noticed while resolving a the merge conflict)

rfk:
A brief update here: I've rebased this atop latest `main`, but I haven't thought deeply about whether and how the approach needs to change now that `apply_pending_updates` has landed. If it all seems fine and good then please feel free to merge this as-is, otherwise I'll dive back in once I'm back from PTO and see what needs to change.

rfk:
>  I haven't thought deeply about whether and how the approach needs to change now that apply_pending_updates has landed

I thought more deeply about this, and I think my position is unchanged from [this earlier comment](https://github.com/mozilla/nimbus-sdk/pull/80#issuecomment-745756536) - the usage of this method is entirely analogous to `set_global_user_participation`, which directly changes experiment state rather than going through `apply_pending_updates`, so we shouldn't try to do any clever pending-updates stuff here either.

Calling this ready for merge.

rfk:
This conflicts pretty badly with https://github.com/mozilla/nimbus-sdk/pull/86, going to back it out

PR #82, opened 2020-12-22
Add enrollment_id to the `EnrolledExperiment` struct

travis79:
In order to have access to the enrollment_id for the purpose of emitting the exposure event from the Android-Component layer, we need to add it to the EnrolledExperiment struct as a string, which already gets passed across the FFI and is available through the get_active_experiments() function.


rfk (approved):
Makes sense, thanks 👍

PR #83, opened 2020-12-22
[SDK-159] Fix experiment example

travis79:
This fixes the experiment CLI example, specifically the `gen-uuid` and `brute-force` commands which didn't quite work after the updates to the `EnrollmentStatus` which added the `enrollment_id`.

### DIFF
--- a/nimbus/examples/experiment.rs
+++ b/nimbus/examples/experiment.rs
@@ -5,12 +5,12 @@
 use clap::{App, Arg, SubCommand};
 use env_logger::Env;
 use nimbus::{
-    error::Result, AppContext, AvailableRandomizationUnits, NimbusClient, RemoteSettingsConfig,
+    error::Result, AppContext, AvailableRandomizationUnits, NimbusClient, RemoteSettingsConfig, EnrollmentStatus
 };
 use std::collections::HashMap;
 use std::io::prelude::*;
 
-const DEFAULT_BASE_URL: &str = "https://settings.stage.mozaws.net"; // TODO: Replace this with prod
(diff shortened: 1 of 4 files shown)


nimbus/examples/experiment.rs
```diff
@@ -324,12 +325,15 @@ fn main() -> Result<()> {
...
                 let uuid = uuid::Uuid::new_v4();
                 let aru = AvailableRandomizationUnits::with_client_id(&client_id);
-                let enrollment =
-                    nimbus::evaluate_enrollment(&uuid, &aru, &Default::default(), &exp)?;
-                results.insert(
-                    enrollment.status.clone(),
-                    results.get(&enrollment.status).unwrap_or(&0) + 1,
-                );
+                let enrollment = nimbus::evaluate_enrollment(&uuid, &aru, &context.clone(), &exp)?;
+                let key = match enrollment.status.clone() {
```
mhammond, line 329: I'd have thought the output before this part of the patch would be better than after, but 🤷

mhammond (changes requested):
Looks fine to me, but clippy isn't happy,

nimbus/examples/experiment.rs
travis79: Because of the addition of the enrollment_id, it was basically categorizing each enrollment by itself with a count of one, so I just simplified what it was categorizing by again.

travis79: It made it difficult to see how many enrolled and didn't enroll

mhammond (approved):

PR #84, opened 2021-01-06
Delete the database if opening it fails with an error indicating it is invalid or corrupt

mhammond:
There are 2 RKV errors that seem to indicate a clearly invalid DB (and [SDK-163](https://jira.mozilla.com/browse/SDK-163) hits one of them.) On these errors we delete the entire `db` directory then recreate it and recreate the DB. This seems a little scary, but the only other alternative is to have a set of files we know to delete, but that seems fragile - there might be other rkv files that only exist in edge-cases (or in future versions) and leaving them behind might cause further problems.

There's a bonus commit which adds an experiment without a schema-version to some tests, but as expected, this is handled correctly. I didn't think it was worth its own PR, but also didn't want to discard it.

### DIFF
--- a/nimbus/src/persistence.rs
+++ b/nimbus/src/persistence.rs
@@ -11,7 +11,7 @@ use crate::error::{Error, Result};
 // production/release environments at Mozilla, you may do so with the "SafeMode"
 // backend", so we really should get more guidance here.)
 use core::iter::Iterator;
-use rkv::StoreOptions;
+use rkv::{StoreError, StoreOptions};
 use std::fs;
 use std::path::Path;
 
@@ -212,7 +212,31 @@ impl Database {
         let path = std::path::Path::new(path.as_ref()).join("db");
         log::debug!("Database path: {:?}", path.display());
         fs::create_dir_all(&path)?;
-        let rkv = rkv_new(&path)?;
+        let rkv = match rkv_new(&path) {
+            Ok(rkv) => Ok(rkv),
+            Err(rkv_error) => {
+                match rkv_error {
+                    // For some errors we just delete the DB and start again.
+                    StoreError::DatabaseCorrupted | StoreError::FileInvalid => {
+                        // On one hand this seems a little dangerous, but on
+                        // the other hand avoids us knowing about the
+                        // underlying implementation (ie, how do we know what
+                        // files might exist in all cases?)
+                        log::warn!(
+                            "Database at '{}' appears corrupt - removing and recreating",
+                            path.display()
+                        );
+                        fs::remove_dir_all(&path)?;
+                        fs::create_dir_all(&path)?;
+                        // TODO: Once we have glean integration we want to
+                        // record telemetry here.
+                        rkv_new(&path)
+                    }
+                    // All other errors are fatal.
+                    _ => Err(rkv_error),
+                }
+            }
+        }?;
         log::debug!("Database initialized");
         Ok(rkv)
     }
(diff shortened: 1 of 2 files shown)


nimbus/src/persistence.rs
```diff
@@ -212,7 +212,29 @@ impl Database {
...
+                        // the other hand avoids us knowing about the
+                        // underlying implementation (ie, how do we know what
+                        // files might exist in all cases?)
+                        log::warn!(
+                            "Database at '{}' appears corrupt - removing and recreating",
+                            path.display()
+                        );
+                        fs::remove_dir_all(&path)?;
+                        fs::create_dir_all(&path)?;
+                        rkv_new(&path)
```
travis79, line 231: What happens if this also fails?

travis79 (commented):
I'm thinking about the case here where we have an enrolled client and somehow the database gets corrupted.  In that case we drop it and create a new one, which would affect the nimbus_id and potentially the enrollment in an ongoing experiment.  If they were part of the treatment branch, it might get messy trying to clean up after the experiment if we no longer were certain that they had the treatment applied or not. I'm not sure what we could do about this other than inform the consuming application.  What do you think?

nimbus/src/persistence.rs
travis79: I suppose it just returns the result then...

mhammond: Yeah, the caller will see that failure.

mhammond:
>  If they were part of the treatment branch, it might get messy trying to clean up after the experiment if we no longer were certain that they had the treatment applied or not.

Yeah, I don't really have an opinion here. Another option is that the a-c side of the world handles this error, but the same problem exists there - but at least in that scenario the app has a way of learning we recovered in that way, but with my patch it does not. I guess I'm hoping that this happens rarely enough that it doesn't matter in practice. Let's raise this at the checkin.

travis79 (approved):
Per the discussion in the meeting, and after checking with DS about this, I think I'm on board with just deleting the database and creating a fresh one.  Let's leave a reminder to record this in telemetry as an error once we move to Glean-Rust.

nimbus/src/persistence.rs
```diff
@@ -212,7 +212,29 @@ impl Database {
...
         log::debug!("Database path: {:?}", path.display());
         fs::create_dir_all(&path)?;
-        let rkv = rkv_new(&path)?;
+        let rkv = match rkv_new(&path) {
+            Ok(rkv) => Ok(rkv),
+            Err(rkv_error) => {
+                match rkv_error {
+                    // For some errors we just delete the DB and start again.
+                    StoreError::DatabaseCorrupted | StoreError::FileInvalid => {
+                        // On one had this seems a little dangerous, but on
```
jhugman, line 221: nit: on the one _hand_.

nimbus/src/persistence.rs
```diff
@@ -321,6 +343,29 @@ mod tests {
...
+        let db_file = db_dir.join("data.safe.bin");
+        #[cfg(not(feature = "rkv-safe-mode"))]
+        let db_file = db_dir.join("data.mdb");
+
+        fs::write(&db_file, b"Not a database!")?;
+        assert_eq!(fs::metadata(&db_file)?.len(), 15);
+        // Opening the DB should delete the corrupt file and replace it.
+        Database::new(&tmp_dir)?;
+        // Old contents should be removed and replaced with actual data.
+        assert_ne!(fs::metadata(&db_file)?.len(), 15);
```
jhugman, line 366: What is the significance of `15` here?

Why would `fs::metadata(&db_file)?` be `b"Not a database!"` in the first instance?

Is this `assert_ne!` proving that we've now got a real database?

Not saying it's wrong, I'm just confused.

jhugman (commented):
I have questions, but they may not result in code changes.

nimbus/src/persistence.rs
mhammond: > What is the significance of `15` here?

That's the length of "Not a database!" - I'll make that more explicit.

> Why would `fs::metadata(&db_file)?` be `b"Not a database!"` in the first instance?

Line 361 writes that - so we expect to see that until the DB has been deleted and recreated.

> Is this `assert_ne!` proving that we've now got a real database?

Yes - I'm not sure how to otherwise check that - I guess I could just check rkv can open it, but that seems easier and just as effective.

mhammond: FTR, I stuck with the same strategy but replaced the magic number with a `.len()` call

PR #85, opened 2021-01-12
Add `get_experiment_branches` method

gabrielluong:
For https://jira.mozilla.com/browse/SDK-101

### DIFF
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -72,6 +72,15 @@ impl NimbusClient {
             .map(|e| e.branch_slug.clone()))
     }
 
+    pub fn get_experiment_branches(&self, slug: String) -> Result<Vec<Branch>> {
+        Ok(self
+            .get_all_experiments()?
+            .iter()
+            .find(|e| e.slug == slug)
+            .map(|e| e.branches.clone())
+            .ok_or(Error::NoSuchExperiment(slug))?)
+    }
+
     pub fn get_global_user_participation(&self) -> Result<bool> {
         // This is a bit smelly, but get_global_user_participation() needs a
         // writer so that the implementation of update_enrollments can pass one
--- a/nimbus/src/nimbus.idl
+++ b/nimbus/src/nimbus.idl
@@ -21,6 +21,17 @@ dictionary EnrolledExperiment {
     string enrollment_id;
 };
 
+dictionary Branch {
+    string slug;
+    u32 ratio;
+    FeatureConfig? feature;
+};
+
+dictionary FeatureConfig {
+    string feature_id;
+    boolean enabled;
+};
+
 dictionary RemoteSettingsConfig {
     string server_url;
     string bucket_name;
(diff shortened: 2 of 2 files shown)


nimbus/src/nimbus.idl
```diff
@@ -19,8 +19,14 @@ dictionary EnrolledExperiment {
     string user_facing_description;
     string branch_slug;
     string enrollment_id;
+    sequence<Branch> branches;
 };
 
+dictionary Branch {
+    string slug;
+    u32 ratio;
+}
```
travis79, line 28: This is missing a semi-colon and is the root cause of the error you are seeing.  See here for more info: https://github.com/mozilla/uniffi-rs/issues/337

nimbus/src/nimbus.idl
```diff
@@ -19,8 +19,14 @@ dictionary EnrolledExperiment {
     string user_facing_description;
     string branch_slug;
     string enrollment_id;
+    sequence<Branch> branches;
 };
 
+dictionary Branch {
+    string slug;
+    u32 ratio;
```
travis79, line 27: This won't work without the missing `feature` definition.  It has to match the one defined in lib.rs exactly, which means you will also need to add the `FeatureConfig` dictionary definition to the IDL.

nimbus/src/lib.rs
```diff
@@ -72,6 +72,14 @@ impl NimbusClient {
             .map(|e| e.branch_slug.clone()))
     }
 
+    pub fn get_experiment_branches(&self, slug: String) -> Result<Vec<Branch>> {
+        Ok(self
+            .get_active_experiments()?
+            .iter()
+            .find(|e| e.slug == slug)
+            .map(|e| e.branches.clone()))
```
travis79, line 80: Since the call to `get_active_experiments()` returns an optional value and has the `?`, you may need to unwrap this to get at the `Vec<Branch>` object.

travis79 (changes requested):

travis79:
After thinking about the approach to this, could you maybe simplify things by not adding the branches to the enrolled experiment struct?  If you instead exposed the `get_experiment_branches` as a method of the nimbus client, it could still return a `Vec<Branch>`.  I'm just not certain of the value of adding it to the `EnrolledExperiment` struct rather than exposing it as a function.

gabrielluong:
> After thinking about the approach to this, could you maybe simplify things by not adding the branches to the enrolled experiment struct? If you instead exposed the `get_experiment_branches` as a method of the nimbus client, it could still return a `Vec<Branch>`. I'm just not certain of the value of adding it to the `EnrolledExperiment` struct rather than exposing it as a function.

Didn't read this in my latest revision. Will think about it and try it for tomorrow.

jhugman (commented):
Drive by, after `f?` in slack.

travis79 (approved):
This looks good to me now!

gabrielluong:
I don't have any write access on this repo. Please land it if it's all good. Thank you!

PR #86, opened 2021-01-13
Make NimbusClient Send+Sync via mutexes and cells

mhammond:
I think this is roughly what will need to happen once uniffi allows us to opt-out of it managing locking

### DIFF
--- a/nimbus/src/dbcache.rs
+++ b/nimbus/src/dbcache.rs
@@ -1,0 +1,89 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::enrollment::get_enrollments;
+use crate::error::{Error, Result};
+use crate::persistence::Database;
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::sync::Mutex;
+
+// This module manages an in-memory cache of the database, so that some
+// functions exposed by nimbus can return results without blocking on any
+// IO. Consumers are expected to call our public `update()` function whenever
+// the database might have changed.
+
+// This struct is the cached data. This is never mutated, but instead
+// recreated every time the cache is updated.
+struct CachedData {
+    pub experiment_branches: HashMap<String, String>,
+}
+
+// This is the public cache API. Each NimbusClient can create one of these and
+// it lives as long as the client - it encapsulates the cells and locking
+// needed to allow the cache to work correctly.
+// WARNING: because this manages locking, the callers of this need to be
+// careful regarding deadlocks - if the callers take their own locks (which
+// they will!), there's always a risk of locks being taken in an inconsistent
+// order. However, there's nothing this code specifically can do about that.
+#[derive(Default)]
+pub struct DatabaseCache {
+    // Notes about the types here:
+    // * We use a `Mutex` because a `RwLock` doesn't make the inner object
+    //   `Sync`, which we require. An alternative would be
+    //   `RwLock<AtomicRefCell<...>>` but we don't have an existing dependency
+    //   on AtomicRefCell and it seems wierd to add one just to micro-optimize
+    //   away from `Mutex` when our uses-cases, in practice, fine with a mutex.
+    //   However, it is worth noting that mozilla-central does depend on
+    //   `AtomicRefCell` so it wouldn't be *that* difficult to argue for the
+    //   new dependency, it's just that no one has yet :)
+    // * We use a `RefCell` even though we don't mutate it in place,
+    //   because `Cell::get()` requires the data to be copied, which a HashMap
+    //   doesn't offer.
+    data: Mutex<RefCell<Option<CachedData>>>,
+}
+
+impl DatabaseCache {
+    // Call this function whenever it's possible that anything cached by this
+    // struct (eg, our enrollments) might have changed. It is passed a
+    // &Database, which implies some mutex guarding that Database is held.
+    pub fn update(&self, db: &Database) -> Result<()> {
+        let experiments = get_enrollments(&db)?;
+        // Build the new hashmap.
+        let mut eb = HashMap::with_capacity(experiments.len());
+        for e in experiments {
+            eb.insert(e.slug, e.branch_slug);
+        }
+        let data = CachedData {
+            experiment_branches: eb,
+        };
+        // then swap it in.
+        let cell = self.data.lock().unwrap();
+        cell.replace(Some(data));
+        Ok(())
+    }
+
+    // Abstracts safely referencing our cached data.
+    fn get_data<T, F>(&self, func: F) -> Result<T>
+    where
+        F: FnOnce(&CachedData) -> T,
+    {
+        let guard = self.data.lock().unwrap();
+        let r = guard.borrow();
+        let data = r.as_ref();
+        match data {
+            None => {
+                log::warn!(
+                    "DatabaseCache attempting to read data before initialization is completed"
+                );
+                Err(Error::DatabaseNotReady)
+            }
+            Some(data) => Ok(func(data)),
+        }
+    }
+
+    pub fn get_experiment_branch(&self, slug: &str) -> Result<Option<String>> {
+        self.get_data(|data| data.experiment_branches.get(slug).cloned())
+    }
+}
--- a/nimbus/src/error.rs
+++ b/nimbus/src/error.rs
@@ -49,6 +49,8 @@ pub enum Error {
     NoSuchBranch(String, String),
     #[error("Server asked the client to back off ({0} seconds remaining)")]
     BackoffError(u64),
+    #[error("Initialization of the database is not yet complete")]
+    DatabaseNotReady,
 }
 
 // This can be replaced with #[from] in the enum definition
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -25,9 +25,10 @@ use enrollment::{
 };
 pub use matcher::AppContext;
 use once_cell::sync::OnceCell;
-use persistence::{Database, StoreId};
+use persistence::{Database, StoreId, Writer};
 use serde_derive::*;
 use std::path::PathBuf;
+use std::sync::Mutex;
 use updating::{read_and_remove_pending_experiments, write_pending_experiments};
 use uuid::Uuid;
 
@@ -38,10 +39,10 @@ const DB_KEY_NIMBUS_ID: &str = "nimbus-id";
 /// It should hold all the information needed to communicate a specific user's
 /// experimentation status
 pub struct NimbusClient {
-    settings_client: Box<dyn SettingsClient + Send>,
+    settings_client: Mutex<Box<dyn SettingsClient + Send>>,
     available_randomization_units: AvailableRandomizationUnits,
     app_context: AppContext,
-    db: OnceCell<Database>,
+    db: OnceCell<Mutex<Database>>,
     db_path: PathBuf,
 }
 
@@ -54,7 +55,7 @@ impl NimbusClient {
         config: Option<RemoteSettingsConfig>,
         available_randomization_units: AvailableRandomizationUnits,
     ) -> Result<Self> {
-        let settings_client = create_client(config)?;
+        let settings_client = Mutex::new(create_client(config)?);
         Ok(Self {
             settings_client,
             available_randomization_units,
@@ -85,41 +86,43 @@ impl NimbusClient {
         // This is a bit smelly, but get_global_user_participation() needs a
         // writer so that the implementation of update_enrollments can pass one
         // and see the correct value.
-        let db = self.db()?;
+        let db = self.db()?.lock().unwrap();
         let writer = db.write()?;
-        get_global_user_participation(db, &writer)
+        get_global_user_participation(&db, &writer)
     }
 
     pub fn set_global_user_participation(
         &self,
         user_participating: bool,
     ) -> Result<Vec<EnrollmentChangeEvent>> {
-        let db = self.db()?;
+        let db = self.db()?.lock().unwrap();
         let mut writer = db.write()?;
-        set_global_user_participation(db, &mut writer, user_participating)?;
+        set_global_user_participation(&db, &mut writer, user_participating)?;
 
         let existing_experiments = db
             .get_store(StoreId::Experiments)
             .collect_all::<Experiment>(&writer)?;
         // We pass the existing experiments as "updated experiments"
         // to the evolver.
-        let nimbus_id = self.nimbus_id()?;
+        let nimbus_id = self.read_or_create_nimbus_id(&db, &mut writer)?;
         let evolver = EnrollmentsEvolver::new(
             &nimbus_id,
             &self.available_randomization_units,
             &self.app_context,
         );
-        let events = evolver.evolve_enrollments_in_db(db, &mut writer, &existing_experiments)?;
+        let events = evolver.evolve_enrollments_in_db(&db, &mut writer, &existing_experiments)?;
         writer.commit()?;
         Ok(events)
     }
 
     pub fn get_active_experiments(&self) -> Result<Vec<EnrolledExperiment>> {
-        get_enrollments(self.db()?)
+        let db = self.db()?.lock().unwrap();
+        get_enrollments(&db)
     }
 
     pub fn get_all_experiments(&self) -> Result<Vec<Experiment>> {
-        self.db()?.collect_all(StoreId::Experiments)
+        let db = self.db()?.lock().unwrap();
+        db.collect_all(StoreId::Experiments)
     }
 
     pub fn opt_in_with_branch(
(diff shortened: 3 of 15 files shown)


mhammond:
This now has 2 commits - the original "send + sync", and also the "get_experiment_branch() should not block" - I was going to put the latter up as a separate PR, but github doesn't want to let me because I use a branch in my fork. I'm obviously open to landing them separately, or not - whatever :)

I'm going to ask for "review" now, although it's probably more accurately considered "feedback" - although I think it's complete, it still depends on uniffi supporting the "threadsafe" concept.

nimbus/src/lib.rs
```diff
@@ -193,16 +237,31 @@ impl NimbusClient {
...
+        self.db
+            .get_or_try_init(|| Ok(Mutex::new(Database::new(&self.db_path)?)))
+    }
+
+    // Call this function whenever it's possible that our enrollments might
+    // have changed. It should be held while the database mutex is held, but
+    // after all IO is complete.
+    fn update_cached_experiment_branches(&self, db: &Database) -> Result<()> {
+        let cell = self.cached_experiment_branches.write().unwrap();
+        let experiments = get_enrollments(&db)?;
```
rfk, line 258: IIUC this is doing to hold the cache write lock for the duration of a DB read, which means that *technically* a concurrent call to `get_experiment_branch` could still have to wait for `get_enrollments` to complete. Would it still work if we read the experiments first, and took the write lock *only* for the duration of the `cell.replace()`? I think the fact that we're holding the database mutex means there won't be any concurrent attempts to update the cache, so there's no risk of writing stale data.

nimbus/src/lib.rs
```diff
@@ -38,10 +41,13 @@ const DB_KEY_NIMBUS_ID: &str = "nimbus-id";
...
 pub struct NimbusClient {
-    settings_client: Box<dyn SettingsClient + Send>,
+    settings_client: Mutex<Box<dyn SettingsClient + Send>>,
     available_randomization_units: AvailableRandomizationUnits,
     app_context: AppContext,
-    db: OnceCell<Database>,
+    db: OnceCell<Mutex<Database>>,
+    // An in-memory cache of our enrollments so that we can answer
+    // get_experiment_branch without doing (or waiting for) IO.
+    cached_experiment_branches: RwLock<RefCell<Option<HashMap<String, String>>>>,
```
rfk, line 50: code-organization nano-nit: you could consider pulling this four-levels-of-nesting-deep type out into a little wrapper `CachedExperimentBranches` struct, along with some helper methods for its core update logic. This would keep the methods of the main `NimbusClient` struct more readable and might give you an obvious place to document the design choices behind it.

Not a big deal though.

rfk (commented):
This all looks pretty great to me! I don't think I should give an official r+ because we'll want to get James's feedback and try it out in conjunction with the uniffi changes, but apart from the one nit around lock duration I think this looks good to go.

nimbus/src/lib.rs
mhammond: Nice catch, thanks!

nimbus/src/lib.rs
mhammond: Good idea - I'll have a poke at that next week.

nimbus/src/lib.rs
```diff
@@ -54,22 +60,34 @@ impl NimbusClient {
...
+        let guard = self.cached_experiment_branches.read().unwrap();
+        let r = guard.borrow();
+        let data = r.as_ref();
+        match data {
+            None => {
+                log::warn!("get_experiment_branch called before initialization is completed");
+                Err(Error::InitializeNotYetComplete)
+            }
+            Some(hashmap) => Ok(hashmap.get(&slug).cloned()),
+        }
```
jhugman, line 90: I guess the whole method could become:

```rust
fn get_experiment_branch(&self, slug: String) -> Result<Option<String>> {
    let db = self.db_cache()?;
    db.get_experiment_branch(slug)
}
```

nimbus/src/nimbus.idl
```diff
@@ -66,7 +66,8 @@ enum Error {
     "JSONError", "EvaluationError", "InvalidExpression", "InvalidFraction",
     "TryFromSliceError", "EmptyRatiosError", "OutOfBoundsError","UrlParsingError",
     "RequestError", "ResponseError", "UuidError", "InvalidExperimentFormat",
-    "InvalidPath", "InternalError", "NoSuchExperiment", "NoSuchBranch", "BackoffError"
+    "InvalidPath", "InternalError", "NoSuchExperiment", "NoSuchBranch", "BackoffError",
+    "InitializeNotYetComplete",
```
jhugman, line 70: Noticing without recommendation: we have begun to stop using the `Error` suffix.

nimbus/tests/common/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+                "isEnrollmentPaused":false,
+                "proposedEnrollment":7,
+                "userFacingDescription":"This is a test experiment for diagnostic purposes.",
+                "id":"secure-gold",
+                "last_modified":1_602_197_324_372i64
+            }
+        ]
+    })
+    .to_string()
+}
```
jhugman, line 94: Nice!

nimbus/tests/test_updates.rs
```diff
@@ -82,32 +23,9 @@ mod test {
         .to_string()
     }
```
jhugman, line 24: Could this go in the new fangled spiffy `common` testing module?

nimbus/tests/test_get_exp_branch.rs
```diff
@@ -0,0 +1,67 @@
...
+        Some("control".to_string())
+    );
+
+    client.set_global_user_participation(false)?;
+    assert_eq!(
+        client.get_experiment_branch("secure-gold".to_string())?,
+        None
+    );
+    Ok(())
+}
```
jhugman, line 67: Nice test!

jhugman (commented):
This is looking good; I particularly what you're doing with the tests.

I've made some suggestions around the `HashMap` which I'd like to see, and a couple of naming nits, which I feel less strongly about.

nimbus/src/lib.rs
mhammond: I came up with something similar to what James suggests.

> when we invariably need to expose more methods to the main thread. e.g. `get_all_experiments()`, `get_global_user_participation()`, `get_experiment_branches()`.

While I agree that is inevitable there will be others, we should resist doing so where possible - ie, we should make sure there *really* is a requirement for doing this from the main thread to avoid creeping complexity. For example, I doubt that any of the functions you mention will need that in practice as they are most likely to be used by the app UI and not directly by experiments (but again, agree that there might well be others)

mhammond:
The new tests had to be disabled in non-safe-mode. I also note that this patch doesn't have the .idl change necessary to actually make this thread-safe. So either:

* This can land before the new uniffi release, then another PR actually marks it as threadsafe
or:
* We can change this PR once that release is cut

I don't think it matters in practice.

nimbus/src/dbcache.rs
```diff
@@ -0,0 +1,74 @@
...
+        match data {
+            None => {
+                log::warn!(
+                    "DatabaseCache attempting to read data before initialization is completed"
+                );
+                Err(Error::DatabaseNotReady)
+            }
+            Some(data) => Ok(func(data)),
+        }
+    }
```
jhugman, line 69: 😻

nimbus/src/error.rs
```diff
@@ -49,6 +49,8 @@ pub enum Error {
     NoSuchBranch(String, String),
     #[error("Server asked the client to back off ({0} seconds remaining)")]
     BackoffError(u64),
+    #[error("Initialization of the database is not yet complete")]
+    DatabaseNotReady,
```
jhugman, line 53: 💯

jhugman (approved):
Excellent stuff. I love how the `DatabaseCache` `update` and `get_data` methods keep everything in one place and easily extendable.

I thought about this since the last time I reviewed this, and have to apologise about naming however. 

I suggested `db_cache`, but I don't think it's right: the `self.db()` and `self.database_cache` operate as different layers, with very different APIs.

Either now or soon, we should consider a `ReadableNimbusClient` trait which is implemented by both `NimbusClient` and `DatabaseCache`, and put all the methods that support access from the main thread.

`DatabaseCache` then becomes `NimbusClientCache`, and we get an easy to copy formalism when we decide to add more, and a `ReadableNimbusClient` trait to reason about in other contexts.

Happy not to block on this, especially if you don't think it's a good idea.

mhammond:
> I suggested `db_cache`, but I don't think it's right: the `self.db()` and `self.database_cache` operate as different layers, with very different APIs.
> 
> Either now or soon, we should consider a `ReadableNimbusClient` trait which is implemented by both `NimbusClient` and `DatabaseCache`, and put all the methods that support access from the main thread.
> 
> `DatabaseCache` then becomes `NimbusClientCache`, and we get an easy to copy formalism when we decide to add more, and a `ReadableNimbusClient` trait to reason about in other contexts.
> 
> Happy not to block on this, especially if you don't think it's a good idea.

I think there's a good idea in there, but TBH I think it's a bit premature to be trying to abstract that away. The current code isn't misleading given there's only one thing cached, so I'm going to suggest we just stick with this for now and consider ideas like the above once we have other use-cases.

I'm not merging this yet given we can't actually leverage it until there's a new uniffi release - at which time we will want another PR to use the new version too - but anyone should feel free to hit the "merge" button as soon as all of those ducks are lined up.

jhugman:
I added a `static_assertions::assert_impl_all!` and found:

```
$ cargo build
   Compiling nimbus-sdk v0.6.4 (/Users/jhugman/workspaces/mozilla/fenix-android-components/nimbus-sdk/nimbus)
error[E0277]: `RefCell<std::option::Option<CachedData>>` cannot be shared between threads safely
  --> src/lib.rs:54:1
   |
54 | static_assertions::assert_impl_all!(NimbusClient: Sync, Send);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   | |
   | `RefCell<std::option::Option<CachedData>>` cannot be shared between threads safely
   | required by this bound in `assert_impl_all`
   |
   = help: the trait `Sync` is not implemented for `RefCell<std::option::Option<CachedData>>`
   = note: required because of the requirements on the impl of `Sync` for `RwLock<RefCell<std::option::Option<CachedData>>>`
   = note: required because it appears within the type `DatabaseCache`
   = note: required because it appears within the type `NimbusClient`
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error
```

:(

jhugman (changes requested):
While we're not yet compiling with `Threadsafe` annotations, we should have a `static_assertion` to check that `NimbusClient` is implementing `Sync + Send`.

rfk:
> While we're not yet compiling with Threadsafe annotations, we should have a static_assertion to check that NimbusClient
> is implementing Sync + Send.

FWIW, I think we *should* get that check for free one `Threadsafe` is available for use, but I support adding an explicit assertion about it in the meantime and keeping it as a way to document the intended requirements.

jhugman:
> FWIW, I think we _should_ get that check for free one `Threadsafe` is available for use, 

That's my understanding too.

> but I support adding an explicit assertion about it in the meantime and keeping it as a way to document the intended requirements.

When I reviewed this, I missed that `RefCell<T>` didn't implement `Sync`, apologies to @mhammond for missing that.

nimbus/src/dbcache.rs
```diff
@@ -0,0 +1,90 @@
...
+    //   `Sync`, which we require. An alternative would be
+    //   `RwLock<AtomicRefCell<...>>` but we don't have an existing dependency
+    //   on AtomicRefCell and it seems wierd to add one just to micro-optimize
+    //   away from `Mutex` when our uses-cases, in practice, fine with a mutex.
+    //   However, it is worth noting that mozilla-central does depend on
+    //   `AtomicRefCell` so it wouldn't be *that* difficult to argue for the
+    //   new dependency, it's just that no one has yet :)
+    // * We use a `RefCell` even though we don't mutate it in place,
+    //   because `Cell::get()` requires the data to be copied, which a HashMap
+    //   doesn't offer.
```
rfk, line 43: Thank you for these A+ comments 👍

nimbus/src/dbcache.rs
```diff
@@ -0,0 +1,90 @@
...
+    data: Mutex<RefCell<Option<CachedData>>>,
+}
+
+impl DatabaseCache {
+    // Call this function whenever it's possible that anything cached by this
+    // struct (eg, our enrollments) might have changed. It is passed a
+    // &Database, which implies some mutex guarding that Database is held -
+    // however, all IO that's performed while this lock is held should already
+    // be complete.
```
rfk, line 52: What will happen if said IO is not in fact complete? (My interpretation is that you'll cache stale data, but I'm not entirely sure if that's correct).

nimbus/src/dbcache.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn update(&self, db: &Database) -> Result<()> {
+        let experiments = get_enrollments(&db)?;
+        let mut eb = HashMap::with_capacity(experiments.len());
+        for e in experiments {
+            eb.insert(e.slug, e.branch_slug);
+        }
+        let data = CachedData {
+            experiment_branches: eb,
+        };
+        // And swap it in.
```
rfk, line 62: nit: The "And" here make this feel like it's the second half of a distributed comment about what this method is doing, but I can't find the first half.

nimbus/src/lib.rs
```diff
@@ -85,82 +95,103 @@ impl NimbusClient {
...
+                Ok(result)
+            }
+            Err(e) => Err(e),
+        }
     }
 
     pub fn opt_out(&self, experiment_slug: String) -> Result<Vec<EnrollmentChangeEvent>> {
-        opt_out(self.db()?, &experiment_slug)
+        let db = self.db()?.lock().unwrap();
+        match opt_out(&db, &experiment_slug) {
```
rfk, line 155: Nit: `and_then` might help reduce control-flow noise a bit here:

```
opt_out(&db, &experiment_slug).and_then(|result| {
    self.database_cache.update(&db)?;
    result
})
```

Dunno if having to explicitly pipe through the value of `result` makes it more obscure on balance though. I think you could also use a variable like:

```
let result = opt_out(&db, &experiment_slug)?;
self.database_cache.update(&db)?;
result
```

nimbus/src/lib.rs
```diff
@@ -170,20 +201,29 @@ impl NimbusClient {
...
     }
 
     pub fn nimbus_id(&self) -> Result<Uuid> {
-        let db = self.db()?;
+        let db = self.db()?.lock().unwrap();
         let mut writer = db.write()?;
+        let uuid = self.read_or_create_nimbus_id(&db, &mut writer)?;
+        // We don't know whether we needed to generate and save the uuid, so
+        // we commit just in case - this is hopefully close to a noop in that
+        // case!
```
rfk, line 215: You could return a `(nimbus_id, was_created)` tuple if you really wanted to know...but I agree it should be cheap in practice.

nimbus/src/nimbus.idl
```diff
@@ -66,9 +66,11 @@ enum Error {
     "JSONError", "EvaluationError", "InvalidExpression", "InvalidFraction",
     "TryFromSliceError", "EmptyRatiosError", "OutOfBoundsError","UrlParsingError",
     "RequestError", "ResponseError", "UuidError", "InvalidExperimentFormat",
-    "InvalidPath", "InternalError", "NoSuchExperiment", "NoSuchBranch", "BackoffError"
+    "InvalidPath", "InternalError", "NoSuchExperiment", "NoSuchBranch", "BackoffError",
+    "DatabaseNotReady",
 };
 
+[Threadsafe]
```
rfk, line 73: 🎉

rfk (approved):
This looks good to me, thanks @mhammond!

nimbus/src/dbcache.rs
mhammond: When writing it, I was thinking "or you may cause some other thread to block on this IO", but that's not true (the lock will be released regardless). Caching stale data if it isn't written yet could also obviously happen, but that seems obvious - so I just removed that fragment.

nimbus/src/lib.rs
mhammond: Yeah, I did bounce around a bit here - the `and_then()` returning the original result almost seems to go against what `and_then()` is designed for, and the latter suggestion means we'll try and update the cache even if we failed, which seems like it could be bad - in particular, it might obscure the original error because the cache update seems likely to also fail.

So I left this alone for now, but I don't feel too strongly about it, so let's see if James does :)

nimbus/src/lib.rs
mhammond: yeah, doesn't seem worth the complexity

nimbus/src/lib.rs
rfk: > the latter suggestion means we'll try and update the cache even if we failed

I don't think it will, the `?` will early-return with an error.

mhammond: doh, right, yeah - that's better (and I don't know how I missed that originally!)

nimbus/src/nimbus.idl
```diff
@@ -78,6 +80,16 @@ interface NimbusClient {
...
+    // Initializes the database and caches enough information so that the
+    // non-blocking API functions (eg, `get_experiment_branch()`) can
+    // return accurate results rather than throwing a "not initialized" error.
+    // It's not strictly necessary to call this function - any function that
+    // wants to use the database will implicitly initialize too - this exists
+    // so that the consuming application can have confidence the non-blocking
+    // functions will return data instead of returning an error, and will do
+    // the minimum amount of work to achieve that.
+    [Throws=Error]
+    void initialize();
```
jhugman, line 92: Nit: `initialize()` sounds like it is necessary (your docs make it's clear that it's not).

When should I use this, rather than `applyPendingExperiments()`? Testing is a reasonable answer, btw.

jhugman (approved):
Lovely.

nimbus/src/dbcache.rs
```diff
@@ -0,0 +1,89 @@
...
+
+// This module manages an in-memory cache of the database, so that some
+// functions exposed by nimbus can return results without blocking on any
+// IO. Consumers are expected to call our public `update()` function whenever
+// the database might have changed.
+
+// This struct is the cached data. This is never mutated, but instead
+// recreated every time the cache is updated.
+struct CachedData {
+    pub experiment_branches: HashMap<String, String>,
```
jhugman, line 20: The more I look at it, the more I think we should be able to read `get_active_experiments` from any thread.

nimbus/src/nimbus.idl
mhammond: As the comment says, whenever you want to do the *least amount of IO* which allows `getExperimentBranch()` to give you an answer. `applyPendingExperiments()` does write IO, so it seems quite clear that Fenix would *not* want to call that if it intends using `getExperimentBranch()` early in startup.

I believe suggesting `applyPendingExperiments()` at startup is going to be an anti-pattern for all the "avoid early IO" discussions we've been having.

nimbus/src/dbcache.rs
mhammond: > The more I look at it, the more I think we should be able to read `get_active_experiments` from any thread.

What's the use-case?

mhammond: And just to clarify - you *can* call it from any thread :) The semantics here are really more about "is it exposed as a synchronous function?" and "does it perform IO?". Platforms will choose to not call such functions from the main thread.

While that seems like splitting hairs, I think the distinction is worthwhile. So in practice, a synchronous, IO-performing function can be called by any thread other than the main thread. If it is exposed as an async function, it can be called just fine from the main thread (even though that implies part of the framework dispatches it to a background thread.)

This is analogous to JS on Desktop - this function will be callable from the main thread, but it will return a promise. I doubt there will be a requirement on Desktop that this be callable synchronously from the main thread. I'm interested to know what use-cases on Fenix would make it a requirement there?

dmose: I think this is related to the [Blocking] / [Non-Blocking] annotation discussed elsewhere, so I've filed mozilla/uniffi-rs#378.

jhugman: Hmm. Yes I could've been clearer.

Currently, we're building UI tools to exercise Nimbus: `getActiveExperiments()` and `getExperimentBranches()` may benefit from being called from the main thread without triggering [StrictMode][1].

A pattern we could arrive at is:

1. read synchronously, without using I/O so we can use it from the main thread; 
2. write everything asynchronously.

At the moment, `getExperimentBranch()` is callable without I/O, but `getExperimentBranches()` is not.

The language specific framework on top of `NimbusClient` can make the API more ergonomic by dispatching writes on to worker threads from the main thread. It cannot make the API nicer if it needs to read from the main thread but Rust insists on doing I/O.

[1]: https://develozper.android.com/reference/android/os/StrictMode

nimbus/src/nimbus.idl
jhugman: I've added your new `initialize()` method to https://github.com/mozilla-mobile/android-components/pull/9394 .

> I believe suggesting applyPendingExperiments() at startup is going to be an anti-pattern for all the "avoid early IO" reasons we've been discussing.

I think at that point, we'd want to optimize the `apply_pending_experiments` + `update_cache` combo to calculate the cache before flushing it to disk. We've had a lot of push back from Fenix to avoid enrollment changes within the same session.

PR #89, opened 2021-01-28
Clarify the release process

mhammond:


README.md
```diff
@@ -12,14 +12,21 @@ Before issuing a new release, the `Unreleased Changes` section should be renamed
 We use [cargo-release](https://crates.io/crates/cargo-release) to simplify the release process.
 Steps:
 
-1. Start a new branch for the release:
-    * `git checkout -b release-vX.Y.Z`
-    * `git push -u origin release-vX.Y.Z`
-2. Update `CHANGELOG.md` as noted above, and commit your changes.
-3. Run `cargo release --dry-run -vv [major|minor|patch]` and check that the things
+1. Ensure your local `main` branch is up to date - the process might try and push this to update tags.
+   * `git co main`
```
rfk, line 16: nit: please use the full `git checkout` here for consistency with the commands below.

README.md
```diff
@@ -12,14 +12,21 @@ Before issuing a new release, the `Unreleased Changes` section should be renamed
...
+   * `git remote update`
```
rfk, line 17: I would have expected the `git pull` to do this automatically, but maybe some folks have it configured not to do that by default?

rfk (approved):
Wow, there were a lot of hidden assumptions in the existing docs, thanks for working through and improving them!

PR #92, opened 2021-01-29
Simplify concurrency primitives in `DatabaseCache`.

rfk:
Today I learned that, since Rust's locking primitives provide mutable references, we don't need constructs like `Mutex<RefCell<T>>` and can instead directly use the lock to provide the mutability.


mhammond (approved):
huh - that's much better!

PR #93, opened 2021-01-29
Add `reset_telemetry_identifiers` method.

rfk:
Fixes https://jira.mozilla.com/browse/SDK-136. This is a continuation of #80 which got backed out for conflicting with the threadsafe work. I'll highlight the things that have changed related to #80 in comments below.

---

When the user disables telemetry in the consuming application, we need
to reset any unique identifiers used internally by the SDK. This guards
against accidentally correlating metrics submitted before the opt-out
with metrics that the user may opt-in to submitting again in future.

Consumers are expected to call `reset_telemetry_identifiers` when the user
disables telemetry in order to trigger resetting of the identifiers.
They may also call it when the user re-enables telemetry in order to
provide updated randomization unit values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,18 +1,33 @@
 # Unreleased changes
 
-# 0.7.1 (_2021-01-28_)
 ## What's New
- - Avoid the use of unsigned "experimental" kotlin types.
+
+- A new `reset_telemetry_identifiers` method has been added; consumers should arrange to call
+  this method if the user opts out of telemetry at the application level, in order to avoid
+  accidental tracking of users who disable then re-enable telemetry.
+
+# 0.7.1 (_2021-01-28_)
+
+## What's New
+- Avoid the use of unsigned "experimental" kotlin types.
 
 # 0.7.0 (_2021-01-28_)
+
 ## What's New
- - Split up `NimbusClient.update_experiments()` into a slow `NimbusClient.fetch_experiments()` and a fast `NimbusClient.apply_pending_experiments()` to help apps manage concurrency and mutable state.
- - Add `set_local_experiments(string)`, to help apps, build tooling for tests, and help during startup on first time run.
- - `get_experiment_branch()` no longer performs any IO, nor blocks on any other threads that may be performing IO, making it suitable for being called from the main-thread of apps.
+
+- Split up `NimbusClient.update_experiments()` into a slow `NimbusClient.fetch_experiments()`
+  and a fast `NimbusClient.apply_pending_experiments()` to help apps manage concurrency and
+  mutable state.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,18 +1,33 @@
...
+
+- A new `reset_telemetry_identifiers` method has been added; consumers should arrange to call
+  this method if the user opts out of telemetry at the application level, in order to avoid
+  accidental tracking of users who disable then re-enable telemetry.
+
 # 0.7.1 (_2021-01-28_)
+
 ## What's New
- - Avoid the use of unsigned "experimental" kotlin types.
+- Avoid the use of unsigned "experimental" kotlin types.
```
rfk, line 12: As before, these are formatting consistency changes, I haven't modified any words from older changelog entries.

nimbus/src/lib.rs
```diff
@@ -37,12 +37,21 @@ use uuid::Uuid;
...
 
+// The main `NimbusClient` struct must not expose any methods that make an `&mut self`,
+// in order to be compatible with the uniffi `[Threadsafe]` annotation. This is a helper
+// struct to contain the bits that do actually need to be mutable, so they can be
+// protected by a Mutex.
+#[derive(Default)]
+struct InternalMutableState {
+    available_randomization_units: AvailableRandomizationUnits,
+}
+
```
rfk, line 48: This `InternalMutableState` stuff is new since #80, designed to make this work with the new requiements that we be `Sync` and `Send`. I've opted to create an obvious place for future additions of mutable state rather than risk a proliferation of individual mutexes.

mhammond: It makes me slightly uncomfortable we have 4 mutexes now - I assume you considered refactoring the database cache into the new mutable state abstraction but decided against it for some reason? I don't have any concrete suggestions here though, so it's not going to block an r+...

mhammond (approved):

rfk:
> It makes me slightly uncomfortable we have 4 mutexes now 

Yeah :-/

> I assume you considered refactoring the database cache into the new mutable state abstraction but
> decided against it for some reason?

Yes, I attempted it, but it got away from me a little bit after I discovered that the db doesn't seem to need to be inside the Mutex; I'm planning to follow up on that in a separate PR.

nimbus/src/lib.rs
rfk: Ref https://github.com/mozilla/nimbus-sdk/pull/94 for a followup that removes one of the mutexes.

Actually I misread your comment as being about the database rather than the database cache; I'll also think about whether it makes sense to move the dbcache into here as well, good suggestion!

PR #94, opened 2021-01-29
Refactor DB to remove mutex, use more explicit transactions.

rfk:
This commit started out with the observation that the Rkv `Database`
class is `Sync` and `Send`, so we don't need to wrap it in a separate
mutex in order to coordinate use from multiple threads. One less mutex
should be a nice cleanup, right?

Turns out we were depending on that mutex for guarding against stale
updates in the database caching layer, so I've refactored that to make
it depend instead on the caller passing in an explicit write transaction.

In the process of doing *that*, I discovered that the way we manage
database transactions seems pretty haphazard. Some functions take just
a `&Database` and manage their own transaction, while their peers take
both `&Database` and `&mut Writer`. We even had a read-only function
that was taking a `&mut Writer` purely so that it could sometimes be
called from within an existing transaction (but sometimes not).

This way lies madness.

So, the bulk of this commit is actually me trying to impose a little
order on how we manage database transactions, based on the following
principles:

* Manage transactions at the outermost layer only. All helper functions
  should take an existing transaction handle, and only the top-level
  client that owns the database should be managing the lifecycle of
  transactions.
* Don't take a write transaction if you don't need one. This has
  required a bit of trait fiddliness but I think it's worthwhile
  to accurately communicate intent (and allow the future possibility
  of more read parallelism).

I'm not wedded to these particular principles, but I do feel strongly that
we need *some* organizing principles for how to manage transactions,
so I'm open to discussion here!

I have some other ideas for how we could clean up the db abstractions here,
but I don't want to get too far ahead of myself.

### DIFF
--- a/nimbus/src/dbcache.rs
+++ b/nimbus/src/dbcache.rs
@@ -4,7 +4,7 @@
 
 use crate::enrollment::get_enrollments;
 use crate::error::{Error, Result};
-use crate::persistence::Database;
+use crate::persistence::{Database, Writer};
 use std::collections::HashMap;
 use std::sync::RwLock;
 
@@ -29,10 +29,20 @@ pub struct DatabaseCache {
 
 impl DatabaseCache {
     // Call this function whenever it's possible that anything cached by this
-    // struct (eg, our enrollments) might have changed. It is passed a
-    // &Database, which implies some mutex guarding that Database is held.
-    pub fn update(&self, db: &Database) -> Result<()> {
-        let experiments = get_enrollments(&db)?;
+    // struct (eg, our enrollments) might have changed.
+    //
+    // This function must be passed a `&Database` and a `Writer`, which it
+    // will commit before updating the in-memory cache. This is a slightly weird
+    // API but it helps encorce two important properties:
+    //
+    //  * By requiring a `Writer`, we ensure mutual exclusion of other db writers
+    //    and thus prevent the possibility of caching stale data.
+    //  * By taking ownership of the `Writer`, we ensure that the calling code
+    //    updates the cache after all of its writes have been performed.
+    pub fn commit_and_update(&self, db: &Database, writer: Writer) -> Result<()> {
+        // By passing in the active `writer` we read the state of enrollments
+        // as written by the calling code, before it's committed to the db.
+        let experiments = get_enrollments(&db, &writer)?;
         // Build the new hashmap.
         let mut eb = HashMap::with_capacity(experiments.len());
         for e in experiments {
@@ -41,8 +51,15 @@ impl DatabaseCache {
         let data = CachedData {
             experiment_branches: eb,
         };
-        // then swap it in.
-        self.data.write().unwrap().replace(data);
+        // Try to commit the change to disk and update the cache as close
+        // together in time as possible. This leaves a small window where another
+        // thread could read new data from disk but see old data in the cache,
+        // but that seems benign in practice given the way we use the cache.
+        // The alternative would be to lock the cache while we commit to disk,
+        // and we don't want to risk blocking the main thread.
+        writer.commit()?;
+        let mut cached = self.data.write().unwrap();
+        cached.replace(data);
         Ok(())
     }
 
--- a/nimbus/src/enrollment.rs
+++ b/nimbus/src/enrollment.rs
@@ -1,9 +1,9 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
-use crate::evaluator::evaluate_enrollment;
 use crate::persistence::{Database, StoreId, Writer};
 use crate::{error::Result, Error};
+use crate::{evaluator::evaluate_enrollment, persistence::Readable};
 use crate::{AppContext, AvailableRandomizationUnits, EnrolledExperiment, Experiment};
 
 use ::uuid::Uuid;
@@ -473,8 +473,12 @@ impl EnrollmentStatus {
 }
 
 /// Return information about all enrolled experiments.
-pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
-    let enrollments: Vec<ExperimentEnrollment> = db.collect_all(StoreId::Enrollments)?;
+pub fn get_enrollments<'r>(
+    db: &Database,
+    reader: &'r impl Readable<'r>,
+) -> Result<Vec<EnrolledExperiment>> {
+    let enrollments: Vec<ExperimentEnrollment> =
+        db.get_store(StoreId::Enrollments).collect_all(reader)?;
     let mut result = Vec::with_capacity(enrollments.len());
     for enrollment in enrollments {
         log::debug!("Have enrollment: {:?}", enrollment);
@@ -484,8 +488,9 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
             ..
         } = &enrollment.status
         {
-            if let Some(experiment) =
-                db.get::<Experiment>(StoreId::Experiments, &enrollment.slug)?
+            if let Some(experiment) = db
+                .get_store(StoreId::Experiments)
+                .get::<Experiment, _>(reader, &enrollment.slug)?
             {
                 result.push(EnrolledExperiment {
                     slug: experiment.slug,
@@ -536,9 +541,9 @@ impl<'a> EnrollmentsEvolver<'a> {
         let is_user_participating = get_global_user_participation(db, writer)?;
         let experiments_store = db.get_store(StoreId::Experiments);
         let enrollments_store = db.get_store(StoreId::Enrollments);
-        let existing_experiments = experiments_store.collect_all::<Experiment>(writer)?;
-        let existing_enrollments =
-            enrollments_store.collect_all::<ExperimentEnrollment>(&writer)?;
+        let existing_experiments: Vec<Experiment> = experiments_store.collect_all(writer)?;
+        let existing_enrollments: Vec<ExperimentEnrollment> =
+            enrollments_store.collect_all(writer)?;
         // Calculate the changes.
         let (updated_enrollments, enrollments_change_events) = self.evolve_enrollments(
             is_user_participating,
@@ -704,38 +709,42 @@ pub enum EnrollmentChangeEventType {
 
 pub fn opt_in_with_branch(
     db: &Database,
+    writer: &mut Writer,
     experiment_slug: &str,
     branch: &str,
 ) -> Result<Vec<EnrollmentChangeEvent>> {
-    let mut writer = db.write()?;
     let mut events = vec![];
-    let exp = db
+    let exp: Experiment = db
         .get_store(StoreId::Experiments)
-        .get::<Experiment>(&writer, experiment_slug)?
+        .get(writer, experiment_slug)?
         .ok_or_else(|| Error::NoSuchExperiment(experiment_slug.to_owned()))?;
     let enrollment = ExperimentEnrollment::from_explicit_opt_in(&exp, branch, &mut events)?;
     db.get_store(StoreId::Enrollments)
-        .put(&mut writer, experiment_slug, &enrollment)?;
-    writer.commit()?;
+        .put(writer, experiment_slug, &enrollment)?;
     Ok(events)
 }
 
-pub fn opt_out(db: &Database, experiment_slug: &str) -> Result<Vec<EnrollmentChangeEvent>> {
-    let mut writer = db.write()?;
+pub fn opt_out(
+    db: &Database,
+    writer: &mut Writer,
+    experiment_slug: &str,
+) -> Result<Vec<EnrollmentChangeEvent>> {
     let mut events = vec![];
     let enr_store = db.get_store(StoreId::Enrollments);
-    let existing_enrollment = enr_store
-        .get::<ExperimentEnrollment>(&writer, experiment_slug)?
+    let existing_enrollment: ExperimentEnrollment = enr_store
+        .get(writer, experiment_slug)?
         .ok_or_else(|| Error::NoSuchExperiment(experiment_slug.to_owned()))?;
     let updated_enrollment = existing_enrollment.on_explicit_opt_out(&mut events)?;
-    enr_store.put(&mut writer, experiment_slug, &updated_enrollment)?;
-    writer.commit()?;
+    enr_store.put(writer, experiment_slug, &updated_enrollment)?;
     Ok(events)
 }
 
-pub fn get_global_user_participation(db: &Database, writer: &Writer) -> Result<bool> {
+pub fn get_global_user_participation<'r>(
+    db: &Database,
+    reader: &'r impl Readable<'r>,
+) -> Result<bool> {
     let store = db.get_store(StoreId::Meta);
-    let opted_in = store.get(writer, DB_KEY_GLOBAL_USER_PARTICIPATION)?;
+    let opted_in = store.get(reader, DB_KEY_GLOBAL_USER_PARTICIPATION)?;
     if let Some(opted_in) = opted_in {
         Ok(opted_in)
     } else {
@@ -760,7 +769,7 @@ pub fn reset_telemetry_identifiers(
 ) -> Result<Vec<EnrollmentChangeEvent>> {
     let mut events = vec![];
     let store = db.get_store(StoreId::Enrollments);
-    let enrollments = store.collect_all::<ExperimentEnrollment>(&writer)?;
+    let enrollments: Vec<ExperimentEnrollment> = store.collect_all(writer)?;
     let updated_enrollments = enrollments
         .iter()
         .map(|enrollment| enrollment.reset_telemetry_identifiers(&mut events));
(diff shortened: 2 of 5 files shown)


rfk:
@jhugman @mhammond @travis79 would love your feedback on this, especially re: the opinionated stance on how to deal with transactions. In my experience the short-term convenience of being able to do some things without explicitly managing a transaction, gets outweighed pretty quickly by the growing complexity of maintaining your mental model of where the transaction boundaries are. But I'm just one guy and I'm often wrong so keen to hear other opinions here :-)

nimbus/src/enrollment.rs
```diff
@@ -1,9 +1,9 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
rfk, line 3: This file was the one that seemed the most in need of some principled opinions - the functions in here variously take or do not take a `Writer` depending on how other parts of the code expect to call them. I've made it so that *none* of these functions are responsible for managing their own transactions.

nimbus/src/enrollment.rs
```diff
@@ -473,8 +473,12 @@ impl EnrollmentStatus {
 }
 
 /// Return information about all enrolled experiments.
-pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
-    let enrollments: Vec<ExperimentEnrollment> = db.collect_all(StoreId::Enrollments)?;
+pub fn get_enrollments<'r>(
+    db: &Database,
+    reader: &'r impl Readable<'r>,
```
rfk, line 478: I like that this clearly-read-only function doesn't take a `Writer`, but I don't like that I had to give it an explicit lifetime parameter. My Rust is not strong enough to figure out whether we can avoid it though.

nimbus/src/enrollment.rs
```diff
@@ -484,8 +488,9 @@ pub fn get_enrollments(db: &Database) -> Result<Vec<EnrolledExperiment>> {
             ..
         } = &enrollment.status
         {
-            if let Some(experiment) =
-                db.get::<Experiment>(StoreId::Experiments, &enrollment.slug)?
+            if let Some(experiment) = db
+                .get_store(StoreId::Experiments)
+                .get::<Experiment, _>(reader, &enrollment.slug)?
```
rfk, line 493: Another function-signature annoyance related to the `Readable` trait- `SingleStore::get` now takes two generic type parameters (one for the data type, one for the concrete transaction type implementing `Readable`) so we have to instruct the turbofish to infer the type of the second.

I think this is actually an opportunity for further future cleanup - we *know* that the `StoreId::Experiments` store is always going to want to return items of type `Experiment`, so we should be able to avoid having to turbofish here at all, perhaps by making explicit types like `ExperimentsStore` and `EnrollmentsStore` and putting the type of contained data as an associated type. But anyway, that's for the future.

nimbus/src/enrollment.rs
```diff
@@ -536,9 +541,9 @@ impl<'a> EnrollmentsEvolver<'a> {
         let is_user_participating = get_global_user_participation(db, writer)?;
         let experiments_store = db.get_store(StoreId::Experiments);
         let enrollments_store = db.get_store(StoreId::Enrollments);
-        let existing_experiments = experiments_store.collect_all::<Experiment>(writer)?;
-        let existing_enrollments =
-            enrollments_store.collect_all::<ExperimentEnrollment>(&writer)?;
+        let existing_experiments: Vec<Experiment> = experiments_store.collect_all(writer)?;
+        let existing_enrollments: Vec<ExperimentEnrollment> =
+            enrollments_store.collect_all(writer)?;
```
rfk, line 546: These explicit type annotations are to avoid the two-arg-turbofish issue noted above.

nimbus/src/lib.rs
```diff
@@ -53,7 +53,7 @@ pub struct NimbusClient {
     settings_client: Mutex<Box<dyn SettingsClient + Send>>,
     mutable_state: Mutex<InternalMutableState>,
     app_context: AppContext,
-    db: OnceCell<Mutex<Database>>,
```
rfk, line 56: I just wanted to do this one simple-looking thing 😭

nimbus/src/lib.rs
```diff
@@ -104,25 +106,21 @@ impl NimbusClient {
     }
 
     pub fn get_global_user_participation(&self) -> Result<bool> {
-        // This is a bit smelly, but get_global_user_participation() needs a
-        // writer so that the implementation of update_enrollments can pass one
-        // and see the correct value.
```
rfk, line 109: On the plus side, IMHO this is now significantly less smelly.

nimbus/src/persistence.rs
```diff
@@ -22,13 +22,23 @@ const DB_VERSION: u16 = 1; // Increment and implement a DB migration in `maybe_u
...
+    pub type Reader<'t> = rkv::Reader<LmdbRoTransaction<'t>>;
     pub type Writer<'t> = rkv::Writer<LmdbRwTransaction<'t>>;
+    pub trait Readable<'r>:
+        rkv::Readable<'r, Database = LmdbDatabase, RoCursor = LmdbRoCursor<'r>>
+    {
+    }
+    impl<'r, T: rkv::Readable<'r, Database = LmdbDatabase, RoCursor = LmdbRoCursor<'r>>>
+        Readable<'r> for T
+    {
+    }
```
rfk, line 41: Trait aliases are experimental, this is a hacky little workaround to simulate importing and specializing the `Readable` trait.

mhammond (approved):
This is amazing and a vast improvement. I need to read it more to fully understand it - the trait looks great, but I'll need to gloss over `<'r>` the same way I gloss over `<'de>` :)

A very brief comment on the client about how we are delegating some of our concurrency story to rkv might help, but that's kinda obvious to anyone looking into that, so I've no concrete suggestion here.

Thanks!

rfk:
> I remember previously being frustrated about having to pass the db just to get the store and you couldn't
> just get it from a reader/writer.

Yes!! I think there's opportunity for an additional refactor there so that you only have to pass one thing around, and in fact *very briefly* started down that road before realizing it was way too much to tackle all at once. But, maybe something we can iterate on.

nimbus/src/enrollment.rs
```diff
@@ -846,8 +855,11 @@ mod tests {
         ]
     }
 
-    fn get_experiment_enrollments(db: &Database) -> Result<Vec<ExperimentEnrollment>> {
-        db.collect_all::<ExperimentEnrollment>(StoreId::Enrollments)
+    fn get_experiment_enrollments(
+        db: &Database,
+        writer: &mut Writer,
+    ) -> Result<Vec<ExperimentEnrollment>> {
+        db.get_store(StoreId::Enrollments).collect_all(writer)
```
jhugman, line 862: Why does this require a `writer`?

nimbus/src/dbcache.rs
```diff
@@ -29,10 +29,18 @@ pub struct DatabaseCache {
...
+    // This function must be passed a `&Database` and a `Writer`, which it
+    // will commit before updating the in-memory cache. This is a slightly weird
+    // API but it helps encorce two important properties:
+    //
+    //  * By requiring a `Writer`, we ensure mutual exclusion of other db writers
+    //    and thus prevent the possibility of caching stale data.
+    //  * By taking ownership of the `Writer`, we ensure that the calling code
+    //    updates the cache after all of its writes have been performed.
+    pub fn commit_and_update(&self, db: &Database, writer: Writer) -> Result<()> {
+        let experiments = get_enrollments(&db, &writer)?;
```
jhugman, line 43: I'm not sure I understand yet: the writer hasn't committed; we're getting enrollments. Is this copy of the old data or the not yet committed data? I can see it's the same `writer`: so confused :(

nimbus/src/dbcache.rs
```diff
@@ -41,7 +49,8 @@ impl DatabaseCache {
         let data = CachedData {
             experiment_branches: eb,
         };
-        // then swap it in.
+        // Commit and swap it in.
+        writer.commit()?;
```
jhugman, line 53: Is this when we're flushing to disk? If so, can we `replace` the data before that happens?

nimbus/src/dbcache.rs
rfk: The idea is that by taking the `writer`, we read the state of the database *from the point-of-view of that transaction*, meaning we read and cache the changes written by the caller.

nimbus/src/dbcache.rs
rfk: > Is this when we're flushing to disk?

It is, yes.

> If so, can we replace the data before that happens?

Yes, but we'll need to be careful about what happens if the commit itself fails. I'm imagining something like:

```
// Get write access to the cache
let cache = self.data.write().unwrap();
// Swap it in. Nobody can read this yet, because of the lock above.
let old_data = cache.replace(data);
// Try to flush the changes to the db on disk
match writer.commit() {
  // Phew, that worked out!
  Ok(_) => Ok(()),
  // Quick, roll back the cache changes before anyone can notice!
  Err(e) => {
    if old_data.is_some() {
        cache.replace(old_data);
    } else {
        cache.take();
    }
    Err(e)
  }
}
// Exclusive access to `self.data` goes out of scope here.
```

@jhugman does that make sense, and seem worthwhile?

nimbus/src/enrollment.rs
rfk: You're right, this should take `reader`, good catch!

nimbus/src/dbcache.rs
rfk: Actually I think it can be simpler than that:

* Take the cache lock
* Commit the transaction
* If it succeeds, update the in-memory cache

I think this will have the same effect of preventing other threads from observing the new data on disk while the old data is still in memory.

rfk: I've gone ahead and made the change here, and added some comments to explain what I think it's doing and why I think it's sound.

nimbus/src/dbcache.rs
rfk: (I added a comment to clarify)

nimbus/src/dbcache.rs
jhugman: > 
>   * Take the cache lock
>   * Commit the transaction
>  * If it succeeds, update the in-memory cache

Sorry, I missed this. 

Checking if I understand [`RwLock`][1] sufficiently: reading from the cache from _the main thread is blocked_ while the data is being written to disk?

[1]: https://doc.rust-lang.org/std/sync/struct.RwLock.html

nimbus/src/dbcache.rs
jhugman: Ok, this makes sense here. 

Can I generalize the rule, by replacing (in my head) a getter with a writer is said to be a `precommit_getter`?

nimbus/src/dbcache.rs
jhugman: > @jhugman does that make sense, and seem worthwhile?

I think so. One of the things I love about this design is we can optimize the caching in a single place it an makes a difference everywhere.

jhugman (commented):
I'm sorry @rfk I didn't see this yesterday.

nimbus/src/dbcache.rs
rfk: That sounds right to me, yes - if you see a `get_*` function taking a `Writer` or `&Writer` then it's doing so in order to read the uncommitted state of the transaction.

(This particular case is a bit special in taking *ownership* of the `Writer` rather than a reference, so it's also taking the responsibility for committing the transaction; that doesn't really generalize and I would expect any other instances of this pattern to be taking a `&Writer` instead)

nimbus/src/dbcache.rs
rfk: > reading from the cache from the main thread is blocked while the data is being written to disk?

From any thread, not just the main thread; calling `self.data.write().unwrap()` takes the write lock and gives us exclusive access to the data it protects, if another thread tries to read the cache then its call to `self.data.read().unwrap()` will block until we release the write lock.

So...I guess the main thread could block waiting for the I/O being performed by `writer.commit()` here. If we want to avoid that, I think fundamentally we'll need to allow the main thread the possibility of reading either stale or uncommitted data from the cache. Thoughts?

rfk:
Commenting for top-level status update: this is pending any further discussion of the potential for blocking the main thread noted [here](https://github.com/mozilla/nimbus-sdk/pull/94#discussion_r567100908).

nimbus/src/dbcache.rs
rfk: If we want to avoid the potential for blocking the main thread, my weakly-held preference is to allow the main thread to read stale data from the cache, by reverting to the original ordering of events:

* Commit the changes to the db.
* *[there is potential for other threads to read stale data between these two steps]*
* Take the write lock on the cache.
* *[potential for reading stale data has now ended]*
* Update the cache.
* Release the write lock.

rfk: In slack, @mhammond talked me into making this a more strongly-held preference. We seem highly unlikely to write any code that would be upset about seeing stale data in the cache here in practice, and such code seems likely to be broken in other ways. So, I've reverted to the logic I had here previously (commit the transaction, then lock and update the cache) but added a comment to explain what's going on and why.

jhugman: > We seem highly unlikely to write any code that would be upset about seeing stale data in the cache here in practice, and such code seems likely to be broken in other ways

I _think_ I disagree with both of these points though admittedly I'm not sure who the "we" is.

My motivators in this discussion are:

0. [Not waiting on I/O on the main thread][1].
1. making up-to-date data available to the main thread as quickly as possible after startup.

I keep talking about the main thread; I appreciate that any thread reading from the cache will  be blocked, but the main thread is where the user perceives it— as dropped frames when drawing the UI and/or an unresponsive UI. The Fenix team is _extremely_ sensitive to this: low-end devices with slow disks are common is their user population. At this stage, the Fenix Perf team sees only performance regressions, not the value that nimbus is bringing. 

Second: by using a read only `DatabaseCache`, we have chosen freeing up the main thread and periods where we what is in the cache does not match what is on disk.

1. **Stale data or no data**. If we write to disk before refreshing the cache, we serve stale data or at startup we don't have a cache at all. Frequent worst case: startup experiments get `DatabaseNotReady`. Of the two: I am more concerned about `DatabaseNotReady` on startup problem than stale data.
2. **Uncommitted data**. If we [refresh the cache before writing to disk][2], we have a period of time where disk errors can happen and we lose the enrollment calculation. Rare worst case: we lose the disk (we can't do anything about) or we re-do the enrollment calculation after the next fetch/apply cycle.

What I don't know is that is this an optimization that is worth doing? At this stage, it doesn't add any additional complexity to the implementation, and it's not a fundamental design decision.

Are there better cheaper mitigations for the "`DatabaseNotReady` at startup problem" that we should take before this one, or should we do this now.

If we chose _not_ to do this, we should probably record this in an issue as a potential optimization.

[1]: https://github.com/mozilla/nimbus-sdk/pull/94/files#r567497717
[2]: https://github.com/mozilla/nimbus-sdk/pull/94/files#r567496810

rfk: > I'm not sure who the "we" is.

By "we" here I meant "the nimbus-sdk developers"; this cache is an implementation detail, any code that touches it is code we've deliberately written to use the cache for some specific reason (such as the not-doing-IO-on-the-main-thread thing).

> My motivators in this discussion are:
>
> * Not waiting on I/O on the main thread.
> *making up-to-date data available to the main thread as quickly as possible after startup.

I agree with the importance of both of these.

> by using a read only DatabaseCache, we have chosen freeing up the main thread and periods where
> we what is in the cache does not match what is on disk.

I agree that's what we're choosing (and what we *should* choose) but I would not frame this as a "read-only DatabaseCache", we're still updating it, we're just not blocking any consumers while we attempt to update it atomically with disk IO.

> If we write to disk before refreshing the cache, we serve stale data or at startup we don't have a cache at all.

At startup we will *always* have no cache, since it's just an in-memory hashmap. So I don't think this changes the startup behaviour.

> Are there better cheaper mitigations for the "DatabaseNotReady at startup problem"

I don't think anything in this PR is a mitigation for the "DatabaseNotReady at startup problem". It's not changing when we populate the cache relative to other things that happen, just the mechanics of doing so.

> If we chose not to do this, we should probably record this in an issue as a potential optimization.

In turn, I have to admit that I'm not entirely sure what "this" refers to here - how are you proposing that the current implementation should change?

This conversations feels fruitful and like it should result in some docs improvements around the cache, which I'll try to take on as part of this PR...but I want to make sure I have an accurate mental model of your mental model of the system before doing so.

rfk: James and I discussed this on slack and agreed to land the current behaviour and file an issue for potential future optimization.

rfk:
> this is pending any further discussion of the potential for blocking the main thread

Top-level comment for visibility: the implementation no longer blocks the main thread on disk IO.

nimbus/src/dbcache.rs
jhugman: And here is that issue:  https://jira.mozilla.com/browse/SDK-189

Thanks, @rfk

PR #95, opened 2021-02-04
Merge v0.7.2 to main

rfk:
This release makes the new `reset_telemetry_identifiers` method available to considers.

This seems to want to merge the release commits from previous `v0.7.x` releases into main as well, although the changelog entries etc are already there. I guess maybe we did a rebase-merge to get them onto `main` rather than creating a merge commit? Anyway, I did a merge commit to try to sync things up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Unreleased changes
+
+# 0.7.0 (_2021-01-28_)
 ## What's New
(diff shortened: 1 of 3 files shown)


travis79 (approved):

rfk:
>  I guess maybe we did a rebase-merge to get them onto main rather than creating a merge commit? 

Ah yes, this repo is configured to prevent merge commits by default. I don't think it's appropriate for releases so I made an executive decision to override it for this PR.

PR #96, opened 2021-02-05
Make get_experiment_branch() search by feature_id and experiment slug

dmose:
Should be more or less set.  New and existing tests are passing.


nimbus/src/nimbus.idl
```diff
@@ -101,6 +102,16 @@ interface NimbusClient {
...
+    // Returns the branch allocated for a given experiment ID. Returns null
+    // if the user is not enrolled in the specified experiment or if the
+    // specified experiment does not exist.
+    [Throws=Error]
+    string? get_branch_slug_for_feature(string feature_id);
+
+    // Returns a list of experiment branches for a given experiment ID.
+    [Throws=Error]
+    sequence<Branch> get_branches_for_feature(string featured_id);
+
```
dmose, line 114: Whoops, gotta update these comments too.

nimbus/src/enrollment.rs
```diff
@@ -412,6 +412,7 @@ pub enum EnrollmentStatus {
         enrollment_id: Uuid, // Random ID used for telemetry events correlation.
         reason: EnrolledReason,
         branch: String,
+        feature_id: String
```
mhammond, line 415: nit: trailing `,` will make future diffs cleaner :)

nimbus/src/enrollment.rs
```diff
@@ -110,7 +110,7 @@ impl ExperimentEnrollment {
         }
         let enrollment = Self {
             slug: experiment.slug.clone(),
-            status: EnrollmentStatus::new_enrolled(EnrolledReason::OptIn, branch_slug),
+            status: EnrollmentStatus::new_enrolled(EnrolledReason::OptIn, branch_slug, &experiment.feature_ids[0],)
```
mhammond, line 113: nit: slightly odd trailing `,` here - I think it should move out of the parens (although you could also argue it should be in *both* places, but IMO that would also be odd, so 🤷 )

nimbus/src/dbcache.rs
```diff
@@ -43,14 +44,21 @@ impl DatabaseCache {
         // By passing in the active `writer` we read the state of enrollments
         // as written by the calling code, before it's committed to the db.
         let experiments = get_enrollments(&db, &writer)?;
+
         // Build the new hashmap.
         let mut eb = HashMap::with_capacity(experiments.len());
```
mhammond, line 49: I like the renaming you did, but I think this var should be renamed too (`eb` made sense when `experiment_branches` existed, but not-so-much now)

mhammond (commented):
Extremely quick drive-by from someone who hasn't read the spec for featureIds in any detail :)

nimbus/src/lib.rs
```diff
@@ -292,6 +307,7 @@ impl NimbusClient {
 
 #[derive(Debug, Clone)]
 pub struct EnrolledExperiment {
+    pub feature_id: String,
```
jhugman, line 310: @travis79 may have a better idea: I thought the mapping from `Experiment` to `EnrolledExperiment` was `at most one`.

Is `EnrolledExperiment` a feature within an experiment, or should this be an array?

jhugman (commented):
Quick drive by 🚚 review. I think most of the trickiness is making sure that each feature gets a maximum of one enrollment.

nimbus/src/dbcache.rs
```diff
@@ -16,7 +16,8 @@ use std::sync::RwLock;
 // This struct is the cached data. This is never mutated, but instead
 // recreated every time the cache is updated.
 struct CachedData {
-    pub experiment_branches: HashMap<String, String>,
+    pub branches_by_exp: HashMap<String, String>,
```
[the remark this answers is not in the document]
dmose: Sold.

nimbus/src/client/http_client.rs
```diff
@@ -227,6 +236,7 @@ mod tests {
                 "proposedEnrollment": 7,
                 "referenceBranch": "control",
                 "probeSets": [],
+                "featureIds": ["some_switch"],
```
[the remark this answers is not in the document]
dmose: My inference from brief discussions with Kate is that this is not currently supported, the reason the schema has already been updated to use an array is more along the lines of future-proofing.

I don't have a clear line-of-sight on how multiple features would work either.  A big part of that is due to me being new to the SDK codebase, but I am also a bit concerned that the details of making multiple features work may have hard-to-resolve complexity.

All that said, I don't have any plans of attempting to address multiple-feature support/enrollment in this PR at all (or even anytime soon).  I've added the array to match the existing schema being output by experimenter, but if we wanted to change that, I suspect it's not too late.

@k88hudson, @travis79 any thoughts here?

nimbus/src/dbcache.rs
```diff
@@ -43,14 +44,21 @@ impl DatabaseCache {
...
         let experiments = get_enrollments(&db, &writer)?;
+
         // Build the new hashmap.
         let mut eb = HashMap::with_capacity(experiments.len());
+        let mut exp_features = HashMap::with_capacity(experiments.len());
+
         for e in experiments {
-            eb.insert(e.slug, e.branch_slug);
+            eb.insert(e.slug, e.branch_slug.clone());
+            exp_features.insert(e.feature_id, e.branch_slug);
```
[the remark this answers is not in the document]
dmose: Yes, agreed.  I'm planning to do that next after this ticket. (This ticket is intended to represent the MVP of something that would support an A/B production experiment).

@jughman I think the acceptance criteria for SDK-190 capture this requirement; I'd love your review of those criteria, however...

nimbus/src/enrollment.rs
dmose: I couldn't find any documentation in this repo on what our policy / usage theory is for `cargo fmt` in this project.  Can someone describe the policy in a way that we could stuff into `README.md`?

nimbus/src/dbcache.rs
```diff
@@ -85,6 +93,10 @@ impl DatabaseCache {
     }
 
     pub fn get_experiment_branch(&self, slug: &str) -> Result<Option<String>> {
-        self.get_data(|data| data.experiment_branches.get(slug).cloned())
+        self.get_data(|data| data.branches_by_exp.get(slug).cloned())
+    }
+
+    pub fn get_branch_slug_by_feature(&self, feature_id: &str) -> Result<Option<String>> {
+        self.get_data(|data| data.branches_by_feature.get(feature_id).cloned())
```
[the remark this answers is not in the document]
dmose: @jhugman I like this in principle, but I'm wondering about the pragmatics of this.  Right now, https://github.com/mozilla-mobile/fenix/search?q=getExperimentBranch returns one invocation and two comments. I was under the impression that we just want to land this PR as soon as possible so that we get as much as we can (eg hopefully at least the SDK-190 edge case checking work).

Would doing the overloaded `get_experiment_branch` idea make SDK-187 (android-components) and SDK-188 (fenix) non-trivially easier?

Anyone else have thoughts here?

nimbus/src/enrollment.rs
dmose: Nevermind -- I see that `cargo fmt` is in the CI!

nimbus/src/dbcache.rs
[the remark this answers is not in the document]
jhugman: My thought process: we either change the Rust here, and [some string changes in the Fenix codebase][1].

Or we could change Rust here, and write kotlin in android components and fenix.

The delays between updating Nimbus, then megazord, then Android Components, then Fenix is the same, but the sequencing is much less complicated.

[1]: https://github.com/mozilla-mobile/fenix/blob/0b764fd3e4eb40b3fc3bf569a9de70a75804a3f1/app/src/main/java/org/mozilla/fenix/experiments/Experiments.kt#L10

nimbus/src/enrollment.rs
dmose: OK, I just ran cargo fmt, which took care of this.

nimbus/src/enrollment.rs
dmose: `cargo fmt` did the trick.

nimbus/src/dbcache.rs
dmose: Done.

nimbus/src/dbcache.rs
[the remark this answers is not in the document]
dmose: Fixed.

nimbus/src/client/http_client.rs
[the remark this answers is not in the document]
travis79: While it does complicate things to have to check that only one experiment is active per feature, one of the original goals of Nimbus was to enable multiple experiments to be running at the same time.  Rather than limiting this to one experiment per client (which I agree, would make things a lot easier), we really do need to ensure that there is only one experiment per feature, but whether or not this happens in the SDK or somewhere before the experiment is launched, I am not for certain.

nimbus/src/dbcache.rs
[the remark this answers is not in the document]
travis79: I had to think on this a minute, but I do like what @jhugman is suggesting as is is less of a breaking change.  `get_experiment_branch(identifier)`, where `identifier` could be either an experiment slug or a feature-id and then internally checks both caches for the branch would keep the API the same on the surface and all we would need to do in Fenix then is to get the right version of Nimbus and change the strings, as James suggests, and we don't have to change the Android Component piece at all (at least if/when we implement the feature API).

nimbus/src/client/http_client.rs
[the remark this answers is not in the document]
k88hudson: > It is significantly easier if only one experiment is active per user.

Yeah, definitely need to allow more than 1 experiment per client.

Re: whether we enforce at the server v.s. the client, ideally the server – the only exception is that sometimes you want to roll-over and re-target the unenrolled users same bucket range if the experiments have different targeting (e.g. one is only first run users, one is any fxa user or something like that)

nimbus/src/lib.rs
```diff
@@ -105,6 +105,20 @@ impl NimbusClient {
             .ok_or(Error::NoSuchExperiment(slug))?)
     }
 
+    // Note: the contract for this function is that it never blocks on IO.
+    pub fn get_branch_slug_for_feature(&self, feature_id: String) -> Result<Option<String>> {
+        self.database_cache.get_branch_slug_by_feature(&feature_id)
+    }
+
+    pub fn get_branches_for_feature(&self, feature_id: String) -> Result<Vec<Branch>> {
```
k88hudson, line 113: What's this function used for?

nimbus/src/client/http_client.rs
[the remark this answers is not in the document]
dmose: Thanks for the clarification.  I think this text doesn't directly effect this PR, so marking it as resolved.

nimbus/src/dbcache.rs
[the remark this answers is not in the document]
dmose: OK, implementation updated to use the strategy suggested by @jhugman.

nimbus/src/dbcache.rs
```diff
@@ -16,7 +16,8 @@ use std::sync::RwLock;
 // This struct is the cached data. This is never mutated, but instead
 // recreated every time the cache is updated.
 struct CachedData {
-    pub experiment_branches: HashMap<String, String>,
+    pub branches_by_experiment: HashMap<String, String>,
+    pub branches_by_feature: HashMap<String, String>,
```
jhugman, line 20: Lovely.

nimbus/src/dbcache.rs
```diff
@@ -43,14 +44,21 @@ impl DatabaseCache {
         // By passing in the active `writer` we read the state of enrollments
         // as written by the calling code, before it's committed to the db.
         let experiments = get_enrollments(&db, &writer)?;
-        // Build the new hashmap.
-        let mut eb = HashMap::with_capacity(experiments.len());
+
+        // Build the new hashmaps
```
jhugman, line 48: I think a comment here about `feature_id`s, and that this is pretty temporary; only one experiment per feature (not multiple features per experiment) is implemented here, rather the in the `EnrollmentEvolver`.

nimbus/src/dbcache.rs
```diff
@@ -84,7 +92,10 @@ impl DatabaseCache {
         }
     }
 
-    pub fn get_experiment_branch(&self, slug: &str) -> Result<Option<String>> {
-        self.get_data(|data| data.experiment_branches.get(slug).cloned())
+    pub fn get_experiment_branch(&self, id: &str) -> Result<Option<String>> {
+        self.get_data(|data| match data.branches_by_feature.get(id) {
+            None => data.branches_by_experiment.get(id).cloned(),
+            Some(branch_slug) => Some(branch_slug.to_owned()),
+        })
```
jhugman, line 99: 😻

nimbus/src/enrollment.rs
```diff
@@ -435,8 +440,9 @@ pub enum EnrollmentStatus {
 }
 
 impl EnrollmentStatus {
-    pub fn new_enrolled(reason: EnrolledReason, branch: &str) -> Self {
+    pub fn new_enrolled(reason: EnrolledReason, branch: &str, feature_id: &str) -> Self {
```
jhugman, line 443: This is temporary, isn't it? Perhaps a comment would be useful, or pass in a `&Branch` that the `fn` can dig out the slug and feature ids from the branch.

nimbus/src/lib.rs
travis79: I think that this should be an array, as per discussion in many places, experiments could potentially have multiple feature ids, basically this should be a duplicate of the array in the `Experiment` object.

nimbus/src/nimbus.idl
```diff
@@ -91,11 +92,13 @@ interface NimbusClient {
     [Throws=Error]
     void initialize();
 
-    // Returns the branch allocated for a given experiment ID. Returns null
-    // if the user is not enrolled in the specified experiment or if the
-    // specified experiment does not exist.
+    // Returns the branch allocated for a given feature_id or experiment_slug.
+    // Search first by feature_id, then by experiment_slug, returning the
+    // the first hit. If the user is enrolled neither in an experiement with id
```
travis79, line 97:
```suggestion
    // the first hit. If the user is enrolled neither in an experiment with id
```

travis79 (approved):
Let's go ahead and change the `EnrolledExperiment` feature id to an array now rather than doing it as followup work, other than that, LGTM!

nimbus/src/lib.rs
jhugman: So is there a need to add `feature_id` in `EnrolledExperiment` at all? Either it has all the feature ids or doesn't need any of them?

I don't think this should hold up this PR unless the answer is obvious, @travis79 .

jhugman (approved):
I have a question, but this is looking super.

@travis79 may be the best person to answer that question.

nimbus/src/lib.rs
travis79: Eh, I don't see a need to have the feature ids in the Android Component, now that you mention it, unless we need to record that in telemetry somewhere, but that should be able to be referenced outside of telemetry right now.  Happy to see it omitted here, unless we determine there is a need for it before we migrate the Glean integration to Rust.

dmose: We have to have at least one `feature_id` there, because the DatabaseCache uses EnrolledExperiments internally, and if they don't expose at least one `feature`, we can't build up the HashTable we need to search on.  I'll switch it to an array.

nimbus/src/nimbus.idl
dmose: Fixed.

nimbus/tests/test_get_experiment_branch_by_feature.rs
```diff
@@ -0,0 +1,80 @@
...
+    client.initialize()?;
+    assert_eq!(client.get_experiment_branch("foo".to_string())?, None);
+    Ok(())
+}
+
+#[cfg(feature = "rkv-safe-mode")]
+#[test]
+fn test_enrolled() -> Result<()> {
+    let _ = env_logger::try_init();
+    let client = common::new_test_client("test_before_open")?;
```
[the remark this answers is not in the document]
dmose: Fixed.  I am, indeed, standing on the shoulders of giants.  :-)

nimbus/src/dbcache.rs
dmose: Thanks to @rfk for helping me sort this out!

nimbus/src/dbcache.rs
dmose: Added.

nimbus/src/enrollment.rs
dmose: Duly commented.

PR #99, opened 2021-02-12
Strawfox Feature API: DO NOT LAND

jhugman:
This PR is an compilable (not runnable) implementation of a proposal for the Feature API for android developers.

It shows:

 * the toy interface between Nimbus and the features
 * a wrapper around some JSON (`Variables`)
 * hand-written examples output of a proposed code generation piece.
 * Toy usage of those generated config objects.

The code generator will add complexity to the build process of whichever project wants to use it (GeckoView, Android-Components, Fenix). Part of this exercise was to explore: 

 * can we live without the code generator if necessary?
 * are there any choices to be made now, which will be difficult to back out of down the road?
 * can fenix developers set-up experiments without considering Rust or Uniffi types


android/src/main/java/org/mozilla/experiments/nimbus/VariablesExamples.kt
```diff
@@ -0,0 +1,168 @@
...
+        setIcon(config.icon)
+
+        setPositiveButton(config.positiveButton.text) { _, _ -> /* on click */ }
+        setNegativeButton(config.negativeButton.text) { dialog, _ ->
+            dialog.cancel()
+        }
+        create()
+    }
+
+    dialogBuilder.show()
```
jhugman, line 118: This is the application code. Everything else is to show the feature developers experience.

android/src/main/java/org/mozilla/experiments/nimbus/VariablesExamples.kt
```diff
@@ -0,0 +1,168 @@
...
+        }
+    """.trimIndent()
+
+    val config = nimbus.getFeatureVariables("example1") { EagerConfig1(it) }
+    val lazyConfig = nimbus.getFeatureVariables("example1") { LazyConfig1(it) }
+
+    val button = Button(context).apply {
+        text = config.buttonText
+        setBackgroundColor(config.buttonColor)
+    }
```
jhugman, line 52: This shows an application developer configuring a button.

jhugman (commented):
This could probably do with an example of configuring a feature without the code generated config objects.

android/src/main/java/org/mozilla/experiments/nimbus/VariablesExamples.kt
rfk: To check my understanding - I interpret this as `configureFeature1` and `configureFeature2` containing the code that an application developer would write, and the other classes etc being code that we would (aspirationally) generate from a higher-level feature description or would (perhaps more likely in the short term) have the developer of the feature provide for use by the application code. Is that accurate?

jhugman: Yes, this is accurate.

I've split the code up into fake generated code and feature code that an application developer may write.

My aim here is to establish a reasonable half way house between app developers dealing with raw JSON like animals, and the full magic of a code generated future.

android/src/main/java/org/mozilla/experiments/nimbus/FakeFeatures.kt
```diff
@@ -0,0 +1,154 @@
...
+            "buttonText": "treatment_button_label"
+        }
+    """.trimIndent()
+
+    // config and lazyConfig are ostensibly the same shape, perform the JSON look up
+    // at different times.
+    val config = nimbus.getFeatureVariables("button-feature")
+
+    val button = Button(context).apply {
+        text = config.getText("buttonText") ?: "Ok then"
```
k88hudson, line 37: is the idea here you would want to define inline defaults sometimes, like for very simple features?

android/src/main/java/org/mozilla/experiments/nimbus/FakeGeneratedFeatureConfig.kt
```diff
@@ -0,0 +1,79 @@
...
+        variables.getText("buttonText") ?: "Okay then"
+    )
+}
+
+// This example does its work bit by bit, lazily grabbing the values out of JSON as it goes.
+// If we think that not every key is going to be used, then this will be the less CPU intensive.
+class ButtonFeatureConfig2(
+    private val variables: Variables
+) {
+    val buttonColor: Int by lazy { variables.getColor("buttonColor") ?: Color.BLUE }
```
k88hudson, line 29: Ah perfect. I was going to bring this up as something that's pretty useful if the feature starts getting kind of big (i'm running into this on Desktop right now)

k88hudson (commented):
🍓 🦊 

I love the Codegen stuff, although I wonder if it's gonna be overkill up front? I think it depends on how much people want features to be complex v.s. just boolean flags in practice. 

I didn't see an `isEnabled` method, is your theory that it's not really needed at the moment, or just didn't put it in?

android/src/main/java/org/mozilla/experiments/nimbus/NimbusClientInterface+.kt
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.experiments.nimbus
+
+var exampleJson: String? = null
+
+// Dummy implementations of new Rust methods exposed via the FFI.
+fun NimbusClientInterface.getFeatureVariablesJson(featureId: String) = exampleJson
```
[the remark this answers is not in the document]
jhugman: FWIW: they're two different implementations: `getFeatureVariablesJson` would be used for the "[pass a single JSON string across the FFI"][1] version.

[1]: https://github.com/mozilla/nimbus-sdk/pull/99/files/80a85ef680883317331830a8748a85397529a2c1#diff-a57f430eb17adee1cb612ab80a16b82bb69c42de7b51c94050203f6ade02372cR25

The second is for the ["multiple calls across the FFI"][2] scheme, for [unpacking a variable with a JSON value][3].

[2]: https://github.com/mozilla/nimbus-sdk/pull/99/files/80a85ef680883317331830a8748a85397529a2c1#diff-a57f430eb17adee1cb612ab80a16b82bb69c42de7b51c94050203f6ade02372cR28
[3]: https://github.com/mozilla/nimbus-sdk/pull/99/files/80a85ef680883317331830a8748a85397529a2c1#diff-a57f430eb17adee1cb612ab80a16b82bb69c42de7b51c94050203f6ade02372cR25

android/src/main/java/org/mozilla/experiments/nimbus/NimbusClientInterface+.kt
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jhugman, line 1: This is where the Rust API starts. We mock it out as extension variables to (a dummy) `NimbusClientInterface`.

android/src/main/java/org/mozilla/experiments/nimbus/Variables.kt
```diff
@@ -0,0 +1,132 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jhugman, line 1: This file defines an interface `Variables`, which gives the user a generic view of the config, with some helper methods to make it easy to write Android code with. It also provides two implementations: one a thin wrapper for a JSON object, and the second which makes calls across the FFI.

For nested configurations, the JSON implementation is used for both.

jhugman (commented):
Additional context for @dmose

jhugman:
Now landed as https://github.com/mozilla/application-services/pull/4075 .

PR #100, opened 2021-02-13
Make featureIds optional to keep bookmarks A/B working (SDK-190)

dmose:
Here are the beginnings of what we need.  So far I've got same enrollment tests running on both a feature with featureIds and one without.  The second set is (correctly, if I've done it right) failing.

Next steps are:

   * find an excited volunteer to drive this forward!  Perhaps @travis79 or @jhugman?  I've invited you both as collabators on my fork, where this branch lives...
   * hack on the code until those tests pass
   * write remaining tests so that all the acceptance criteria in [SDK-190](https://jira.mozilla.com/browse/SDK-190) are covered
   * make those tests pass
   * review & land

PR #101, opened 2021-02-15
Add test to prove experiments without a feature_ids field work

jhugman:
This PR replaces #100, and fixes SDK-190.

Not long ago, experimenter was producing experiment JSON without an array for `feature_ids`.

Unfortunately, some of these experiments are still live. The experiments were not being parsed,
and mass enrollments happened.

Going forward, `feature_ids` should be a required field.

This PR defaults `feature_ids` when missing to an empty vec of strings. For parts of the code that require a `featue_id`, an empty string is used. This will need to be revisited once we actually support multiple features per branch.

### DIFF
--- a/nimbus/src/lib.rs
+++ b/nimbus/src/lib.rs
@@ -319,6 +319,7 @@ pub struct Experiment {
     pub bucket_config: BucketConfig,
     pub probe_sets: Vec<String>,
     pub branches: Vec<Branch>,
+    #[serde(default)]
     pub feature_ids: Vec<String>,
     pub targeting: Option<String>,
     pub start_date: Option<String>, // TODO: Use a date format here
@@ -335,6 +336,14 @@ impl Experiment {
         self.branches
             .iter()
             .any(|branch| branch.slug == branch_slug)
+    }
+
+    fn get_first_feature_id(&self) -> String {
+        if self.feature_ids.is_empty() {
+            "".to_string()
+        } else {
+            self.feature_ids[0].clone()
+        }
     }
 }
 
--- a/nimbus/tests/common/mod.rs
+++ b/nimbus/tests/common/mod.rs
@@ -30,6 +30,101 @@ pub fn new_test_client(identifier: &str) -> Result<NimbusClient> {
 }
 
 #[allow(dead_code)] // not clear why this is necessary...
+pub fn exactly_two_experiments() -> String {
+    use serde_json::json;
+    json!({
+        "data": [
+            {
+                "schemaVersion": "1.0.0",
+                "slug": "startup-gold",
+                "endDate": null,
+                "featureIds": ["aboutmonkeys"],
+                "branches":[
+                    {
+                        "slug": "control",
+                        "ratio": 1,
+                        "feature": {
+                            "featureId": "aboutmonkeys",
+                            "enabled": false
+                        }
+                    },
+                    {
+                        "slug": "treatment",
+                        "ratio":1,
+                        "feature": {
+                            "featureId": "aboutmonkeys",
+                            "enabled": true
+                        },
+
+                    }
+                ],
+                "probeSets":[],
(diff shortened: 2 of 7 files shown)


nimbus/src/lib.rs
```diff
@@ -319,6 +319,7 @@ pub struct Experiment {
     pub bucket_config: BucketConfig,
     pub probe_sets: Vec<String>,
     pub branches: Vec<Branch>,
+    #[serde(default)]
```
jhugman, line 322:
```rust
// ⚠️ Warning : Altering this type might require a DB migration. ⚠️
```

This change affects the reading of experiments that are serialized and are missing a `featureIds` field.

The `Experiment` objects in the database either have a `feature_ids` property, in which case, this line has no effect; or not: in which case we default to an empty array.

In this PR we test that this deserialization with and without a `featureIds` field. 

I believe that we do not need to do any migration at this point.

jhugman (commented):
TL;DR added a default of an empty array. There are questions around the `feature` property in `branch`, but this is out of scope for this PR.

nimbus/src/lib.rs
rfk: > I believe that we do not need to do any migration at this point.

+1, your reasoning makes sense to me w.r.t. not needing a db migration.

Longer-term, we could consider adding some explicit tests for reading data in order formats from the persistence layer, which I think is pretty similar to what you're doing here, but structured in a way that makes it more obvious *why* it's important to test these things. The [fxa-client state-persistence tests](https://github.com/mozilla/application-services/blob/3eda75aa501fad5161d8c0538c7b3cdf89da18b9/components/fxa-client/src/state_persistence.rs#L236) might be good inspiration here. (Each of these tests as a comment saying "It's really important that you don't modify the test data here, as that would defeat the point of the test" precisely because we were accidentally putting up PRs that modified the b/w-compat test data :-/)

rfk:
FWIW, I was able to produce the `cargo fmt` changes that are failing this in CI by doing a `rustup update stable` to get on `rustc v1.50.0`. (And I continue to dislike the spurious CI failures we get by not pinning to a specific Rust version, but not enough to attempt to re-litigate that cause...)

nimbus/tests/common/mod.rs
```diff
@@ -124,6 +124,134 @@ pub fn initial_test_experiments() -> String {
...
+                "isEnrollmentPaused":false,
+                "proposedEnrollment":7,
+                "userFacingDescription":"This is a test experiment for diagnostic purposes.",
+                "id":"secure-gold",
+                "last_modified":1_602_197_324_372i64
+            },
+            {
+                "schemaVersion": "1.0.0",
+                "slug": "no-features",
+                "endDate": null,
```
rfk, line 220: There may be some small value in putting a comment here saying "This experiment is deliberately missing the `featureIds` field" or similar.

nimbus/tests/test_get_experiment_branch_by_id.rs
```diff
@@ -26,44 +26,65 @@ fn test_before_open() -> Result<()> {
...
-    assert_eq!(
-        client.get_experiment_branch("secure-gold".to_string())?,
-        None
-    );
-    client.apply_pending_experiments()?;
-    // secure-gold enrolls everyone - not clear what treatment though.
-    assert!(client
-        .get_experiment_branch("secure-gold".to_string())?
-        .is_some());
+    let experiment_slugs = vec!["secure-gold", "no-features"];
```
rfk, line 33: My parse of the diff here is that the previous version of the test did a bunch of opting-in and opting-out with the `secure-gold` experiment, and the new version of the test is now doing the same checks against both `secure-gold` and `no-features`. That sounds good and worthwhile but please correct me if I've dramatically misunderstood the intent here :-)

rfk (approved):
LGTM!

PR #102, opened 2021-02-17
Fill in a default feature_id for enrollments where it is missing.

rfk:
In #96 we added a new `feature_id` field to `EnrollmentStatus::Enrolled` but did not add a corresponding db migration. Since there may be Nimbus databases in the wild with serialized `EnrollmentStatus` values that are missing this field, let's default it to the empty string, to avoid bricking the db with serde deserialization failures.

@jhugman Sorry I missed this in the review of #101 but stumbled across it while preparing to cut a new release. I've tried to add some more explicit tests and docs to help us avoid similar oversights in future.

(More broadly, it's really unfortunate that a single corrupted item in the store currently means that any call to `collect_all()` for that store will fail, basically bricking the store. We should do better, but that's a bigger change for a separate PR)

### DIFF
--- a/nimbus/src/enrollment.rs
+++ b/nimbus/src/enrollment.rs
@@ -18,7 +18,8 @@ const DEFAULT_GLOBAL_USER_PARTICIPATION: bool = true;
 const PREVIOUS_ENROLLMENTS_GC_TIME: Duration = Duration::from_secs(30 * 24 * 3600);
 
 // These are types we use internally for managing enrollments.
-// ⚠️ Warning : Altering this type might require a DB migration. ⚠️
+// ⚠️ Attention : Changes to this type should be accompanied by a new test  ⚠️
+// ⚠️ in `mod test_schema_bw_compat` below, and may require a DB migration. ⚠️
 #[derive(Deserialize, Serialize, Debug, Clone, Hash, Eq, PartialEq)]
 pub enum EnrolledReason {
     Qualified, // A normal enrollment as per the experiment's rules.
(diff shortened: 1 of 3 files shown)


rfk:
> More broadly, it's really unfortunate that a single corrupted item in the store currently means that any call to collect_all()
> for that store will fail, basically bricking the store. 

Some more context on this comment, for reference.

We currently process the list of enrollments [like this](https://github.com/mozilla/nimbus-sdk/blob/e6b96257ae0c48b98f219e4427522731253fe862/nimbus/src/enrollment.rs#L489):

```
let enrollments: Vec<ExperimentEnrollment> = db.get_store(StoreId::Enrollments).collect_all(reader)?;
```

Through the magic of the type system, this delegates to [a little loop inside `collect_all`](https://github.com/mozilla/nimbus-sdk/blob/e6b96257ae0c48b98f219e4427522731253fe862/nimbus/src/persistence.rs#L151) which reads each item from the database, deserializes it, and appends it to a vector:

```
        let mut iter = self.store.iter_start(reader)?;
        while let Some(Ok((_, data))) = iter.next() {
            if let rkv::Value::Json(data) = data {
                result.push(serde_json::from_str::<T>(&data)?);
            }
        }
```

By using the `?` operator on `serde_json::from_str`, we cause any deserialization failure to immediately exit the function and return its error to the caller. But we don't seem to have any code to catch these errors and do anything about them, we just propagate them all the way up to the application code. If we get some unserializable data in one of those keys, then we will not be able to read any enrollments until the data somehow gets cleared from the db.

That's...well, let's charitably call it "fragile" :-)

If we shipped without this PR, I expect that existing users who have enrollment records already in their db would end up stuck with a non-working nimbus, unable to participate in any future experiments because we can't read their existing enrollments from the database. I believe we have observed this at least once in a earlier deploy of Nimbus to Fenix Nightly, which is what caused us to add the existing db migration logic in the code here.

Some options that might be better here:

* Skip individual items that fail to deserialize, rather than erroring out completely.
    * We could just ignore them, or we could delete them from the db.
* Have `collect_all::<T>` return a tuple `(Vec<T>, Result<()>)` giving *both* the list of things we were able to read, and any errors encountered for thins we couldn't.

Any other suggestions? /cc @mhammond @travis79 for additional perspectives here.

travis79 (approved):
The flexibility of serde never disappoints!

PR #106, opened 2021-02-22
Remove unused gradle build config from this repo.

rfk:
Not sure if this will be controversial, but here goes...

All consumers of the Nimbus Rust SDK are expected to get it via the
Application Services framework. In fact, if they try to use a version
of Nimbus that wasn't build as part of the Application Services
"megazord" then it will not actually work, because the networking
layer will not have been properly initialized.

To ease maintenance and avoid future confusion, let's delete the
standalone android publishing config from the nimbus-sdk repo,
since we never expect anyone to consume the artifacts it produces
in practice.

Since we're not producing built artifacts from this repo, there is
no point in checking in the `Cargo.lock` file for the Rust crate,
so that has also been removed.

(If we're happy to go ahead with this, then I think we should also
move the Rust `nimbus` crate up to the top level of the repo, so
that running `cargo` in the root of the checkout works. But that's
for a followup PR).


rfk:
This PR is not complicated but it's pretty opinionated, so I'm asking each of @mhammond @travis79 @jhugman individually for review.

rfk:
(Ref https://github.com/mozilla/application-services/pull/3878 for the issue that inspired me to suggest this, in which the fact that Nimbus pretends to have its own standalone build logic makes certain appservices-side changes harder).

mhammond (approved):
👍 I agree with Ryan's rationale

rfk:
Hrm, I note that @travis79 was trying to use these gradle build scripts as part of the "better builds" prototype over in https://github.com/travis79/better-builds-prototype/pull/1 (although IIUC it wasn't really working out just yet).

travis79 (approved):
Sounds good to me, although I noticed you also deleted the Cargo.lock file?

jhugman (approved):
I have mixed feelings about this.

From a "making Android engineers at Mozilla lives easier" point of view, I think this is the right thing to do. Ultimately, that's why I have r+d.

However; I think the handwritten foreign-language extensions to `NimbusClient` e.g. Features API, should live in a place closer to Nimbus than to the platform-specific, browser-adjacent projects.

I don't want to stand in the way of this PR, but wanted to signal my desire to have a wider conversation about this at some point.

rfk:
> I noticed you also deleted the Cargo.lock file?

Yes, deliberately, because it doesn't serve any useful purpose if we're not producing built artifacts from this repo. (The `Cargo.lock` that counts is the one in application-services which is used when we build the artifacts.)

> However; I think the handwritten foreign-language extensions to NimbusClient e.g. Features API,
> should live in a place closer to Nimbus than to the platform-specific, browser-adjacent projects.

This is a good point, and I broadly agree they should be managed as close to the Rust code as possible.

I think we should go ahead with removing this code in the short term, but leave ourselves the option of bringing back some actually-used android build stuff in the future, perhaps as part of the "better builds" prototyping work.

PR #108, opened 2021-02-24
Rename `nimbus.idl` to `nimbus.udl`.

rfk:
We've settled on `.udl` as a naming convention for these files.
In my work on using UniFFI with the fxa-client crate, I'm adding
some XCode build rules that deal with `.udl` files, so switching
things here to be consistent will help smooth a future iOS integration.


jhugman (approved):
❤️ Thank you.

PR #110, opened 2021-03-09
Rename Error to NimbusError

jhugman:
This PR prevents collisions between `Error` (from Nimbus) and `Error` from Swift.

The Swift `Error` isn't namespaced, so when compiling the generated swift file, the Nimbus Error is mistaken for the Swift one.

This is split into two implementations: the first commit is the minimally invasive change, using import aliasing.

The second does the actual rename. Reviewers are invited to tell me if I should drop the second commit or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## What's Changed
 
 - Upgraded uniffi to 0.8.0. This is to allow nimbus to be includable in the iOS megazord.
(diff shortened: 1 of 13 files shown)


travis79 (approved):
Looks good to me, too bad the import aliasing didn't work out. 

Since everything in the Nimbus android component is namespaced under `org.mozilla.experiments.nimbus.` there shouldn't be any downstream changes on the Android side, right? 🤞

PR #111, opened 2021-03-09
(cargo-release) version 0.8.3

jhugman:


rfk (changes requested):
Thanks @jhugman. The commit contents say it's actually `v0.8.3` so we should probably make sure the commit message and git tag match that.

However, if we're going to have to fix those up anyway, I fee like this should be a `v0.9.0` given that the API exposed by this component has changed in a b/w incompatible way (even though that probably won't affect our actual consumers due to Android API wrapping).

PR #116, opened 2021-04-15
Disabling CircleCI and Codecov and putting up a "We have moved" sign

travis79:


travis79:
Oops, didn't mean to include previously merged commits...

